PR #1, opened 2017-04-12
add 3 more tools

bgruening:


bgruening:
@TorHou you need to update ephemeris for the new global options.

TorHou:
@bgruening I see, I like global options, those dependency parameters were "a bit" redundant. There is no way to test these changes easily right?

PR #3, opened 2017-04-21
Not sure about the section here

bgruening:
@TorHou


TorHou:
I was thinking of a section named "Interval manipulation" or "Annotation manipulation" or something like that. But for now this is fine.

bgruening:
Genomic coordinate manipulation?

TorHou:
Installation worked fine I'll put it on the server.

PR #4, opened 2017-05-08
One tool moved to get rid of a section.

TorHou:
Hi Björn. This should serve as an example of how I would adapt the format in the tools.yaml
Note that I don't use the 

> tool_panel_section_label

 but the 
> tool_panel_section_id

instead. I think that's cleaner. And also I leave out the information about the dependencies if they are as per default. If you think this is fine and accept the pull request, then I will do the other changes accordingly.

### DIFF
--- a/tools.yaml
+++ b/tools.yaml
@@ -4,6 +4,12 @@ install_resolver_dependencies: true
 install_tool_dependencies: false
 
 tools:
+
+- name: dwt_var_perfeature
+  owner: devteam
+  tool_panel_section_id: statistics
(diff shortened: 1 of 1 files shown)


TorHou:
From the bioblend source:
```
        if tool_panel_section_id:
            payload['tool_panel_section_id'] = tool_panel_section_id
        elif new_tool_panel_section_label:
            payload['new_tool_panel_section_label'] = new_tool_panel_section_label
```
So the id is more important than the label. But you can still specify a new label for a tool by leaving out the id.

bgruening:
Ok, let's go with this.

PR #10, opened 2017-05-19
Add mothur and qiime

bgruening:
ping @bebatut @TorHou not sure about the section


TorHou:
:+1: it's good for me. We just have to add the sections under the right label in the tool_conf.xml

bgruening:
@TorHou can you add the label please.

PR #12, opened 2017-06-23
Add rory and gubbins

bgruening:
ping @TorHou


bgruening:
Yes, can not hurt. I'm currently not clear how we should organize this and what is useful. But adding it is good for now.

PR #16, opened 2017-08-04
Add some new tools from W4M - version 3.0

bgruening:
ping @lecorguille


lecorguille:
It is always a bit disturbing this **fr** in galaxy-**fr**-tools since it means France for me 😝 
But no problem with that!
Thanks you for your interest Björn

bgruening:
Upps, we can rename the repo ;) I will do this.

lecorguille:
It was necessary :)

PR #22, opened 2017-08-23
Pin all repositories

erasche:
We're still missing a couple but this should enable us to safely run the "install/update all".

I propose that by default, people add new repositories as unpinned, and then we'll have a jenkins job which pins them after installation. Does this sound fine?


bgruening:
Like this in general but would use this syntax if easily possible. Not strictly needed.
```
changeset_revision:
    - cd253831a014
```

scripts/fix-revision.py
```diff
@@ -0,0 +1,22 @@
+import json
```
bgruening: Can we use something like this https://stackoverflow.com/questions/42518067/how-to-use-ordereddict-as-an-input-in-yaml-dump-or-yaml-safe-dump to keep the ordering. Having changeset before name seems odd.

erasche: done.

erasche:
Ok:

- all yaml files have been reverted to master
- instead we generate lock files and no one needs to care about what these look like
- there is a script, update-trusted.py which will update the lock files with new tools, and for any "trusted" repos it will add the latest version to the installation list.
- this doesn't cover 100% of installed repos, we will handle that at a later date, i want to make an incremental improvement and automate the addition of new tools first. having 100% coverage is not necessary for improving the tool installation process.

bgruening:
Thanks!

PR #29, opened 2017-09-04
Add ASaiM tools

bebatut:


asaim.yaml
```diff
@@ -0,0 +1,338 @@
...
+
+- name: 'collection_column_join'
+  owner: 'iuc'
+  tool_panel_section_label: 'Text Manipulation'
+
+- name: 'column_maker'
+  owner: 'devteam'
+  tool_panel_section_label: 'Text Manipulation'
+
+- name: 'concatenate_multiple_datasets'
```
bgruening: Is this really needed? I think this can be replaced by either the text-processing tools or the new data-collection tools.

bebatut: Which data-collection tools?

bgruening:
please don't merge yet, we need to see if all tools are needed ...
Many of the text manipulating tools can probably be replace by existing once.

erasche:
ok @bgruening. None will be installed, I still haven't fixed the jenkins portion.

erasche:
This has been open for 20 days, I'd love to merge this and resolve the text manipulation tools at a later date. Bearing in mind that *none* of these tools will be installed since the jenkins job is still being re-written.

bgruening:
@erasche there are a few tools included there that can be properly replaced by other already installed tools ... just not entirely sure. Need to look over this with @bebatut

erasche:
@bgruening yeah, I saw the previous discussion, I was just considering that y'all might not have a bunch of time in the next $n months and thought we might merge and fix later.

bgruening:
Cdhit is already installed afaik from a different user. This should not be installed but upstreamed, no?

bgruening:
@bebatut are all convertion tools etc really necessary? Why are these not part of the tools?
I'm just concert of tool that are too specific ...

bebatut:
Which convertion tools? I tried to remove them. Did I miss some?

bgruening:
`format_cd_hit_output` or `format_metaphlan2_output` for example.

PR #30, opened 2017-09-04
Fix categories

bgruening:
Just a start ...

### DIFF
--- a/epigenetics.yaml
+++ b/epigenetics.yaml
@@ -1,6 +1,7 @@
 ---
-install_resolver_dependencies: True
-install_tool_dependencies: False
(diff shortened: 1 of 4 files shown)


erasche:
rebased for new linting.

erasche:
Some new things to fix I'm afraid.

bgruening:
I rather think your linter is not correct ;)
There is also `tool_panel_section_id` which is used a lot. `tool_panel_section_id` is very specific and what Galaxy is using the `tool_panel_section_label` is converted with a strange, but deterministic algorithm to an ID.

erasche:
Ah, I missed that one. Will correct the testing. :)

PR #33, opened 2017-09-26
add flash

bgruening:


bgruening:
Waaa, nice tests, but too much more today to fix! Thanks @erasche!

PR #36, opened 2017-09-29
Pykwalify

erasche:
- Adds a schema
- Validates against that schema
- Removes extraneous use of travis + jenkins for just jenkins (since we have secrets involved)
- all toolshed urls are forced to `https`
- swaps a bunch of section IDs to section Labels


tools.yaml
```diff
@@ -110,768 +110,768 @@ tools:
   - name: ega_download_streamer
     owner: yhoogstrate
     tool_panel_section_label: Get Data
-    tool_shed_url: toolshed.g2.bx.psu.edu
+    tool_shed_url: https://toolshed.g2.bx.psu.edu
```
bgruening: We should remove the toolshed from the sections entirely. We do only support TS and not TTS ... so this should go ontop.

erasche: Shall I remove:

```
- name: covenntree
  owner: steffen
  tool_panel_section_label: Graph/Display Data
  tool_shed_url: https://testtoolshed.g2.bx.psu.edu  
- name: protease_prediction
  owner: bgruening
  tool_shed_url: https://testtoolshed.g2.bx.psu.edu 
- name: jamm
  owner: messersc
  tool_panel_section_label: Peak Calling
  tool_shed_url: https://testtoolshed.g2.bx.psu.edu 
```

bgruening: yes please.
Protease prediction should be on the TS I think.

erasche: cool, thanks.

PR #45, opened 2017-10-31
add new proteomics tools

bgruening:


erasche:
linting errors look unrelated, not sure how we still have those :s

PR #46, opened 2017-11-02
hicexplorer is now also IUC tools

bgruening:


bgruening:
I added minimap2 to this PR.

PR #47, opened 2017-11-02
Add metagenomics assemblers

bebatut:
We could also add them to "Metagenomic analysis" section. What do you think?


bgruening:
Galaxy can not handle multiple sections properly and the search will then show two hits.
:+1: for the tools, if you fix the errors :)

bebatut:
The errors are not on this YAML file but `tools.yaml`. Will fix that tm in an other PR

PR #49, opened 2017-11-04
fix metaspades

bgruening:
ping @bebatut


bgruening:
got it, sorry for the noise

PR #73, opened 2018-01-03
add annotatemyids

bgruening:


erasche:
@bgruening this waiting on anything?

bgruening:
No, but since we do not have the automatic installation in place I often end up installing it manually, adding it here (hopefully most of the times) and forget about it.

erasche:
Soon after I get back :) Will fix. Promise.

PR #76, opened 2018-02-12
Trust everything in IUC file.

erasche:
@bgruening just letting you know how it is implemented currently. Will fix to trust the file as a whole soon.


erasche:
Fixed, now it just by file rather than a horrible list. Any anyone who wants to update a file, just runs the script on that file. (All will be documented soon.)

PR #87, opened 2018-03-08
add MiModD suite

bgruening:


bgruening:
depending on https://github.com/usegalaxy-eu/galaxy-playbook-temporary/pull/193

tools.yaml
```diff
@@ -4,17 +4,21 @@ install_resolver_dependencies: true
 install_tool_dependencies: false
 
 tools:
+  - name: suite_mimodd
```
wm75: mimodd_main

would be better than suite_mimodd (the suite really brings in only stuff that's unnecessary on a well-equipped server; you don't need another aligner, nor another SnpEff wrapper I would say)

erasche:
(merging now, it won't get installed until this weekend)

tools.yaml
bgruening: Mh, that is confusing ;)

bgruening: Changed it on master.

wm75:
After things got installed you may want to have a look at
http://mimodd.readthedocs.io/en/latest/install_toolshed.html#enhancing-the-user-experience .
The display whitelisting would be most important.

In addition, it would be great if you could configure the job runner for the Variant Calling tool to enable
multiprocessing. The tool parallelizes variant calling by contig and since most of our end users are working with C. elegans, which has 6 chromosomes, 6 cores would be an ideal value.

tools.yaml
wm75: perfect, thanks.

PR #95, opened 2018-05-04
Update cd-hit which fix a typo for Pavan

bebatut:


bebatut:
merged by @erasche !!!!!!!

bgruening:
@bebatut can we please get rid of this CD-hit implementation. On IUC are two open PRs with cd-hit migration instructions. On usegalaxy.eu we have now two different cd-hit tools :(

erasche:
@bgruening until the IUC one is merged, we need a functional one to install from the TS. This is it.

bgruening:
> This is it.

What?

The PRs are open since ages. Duplicated tools are a no-go and we should try to avoid it as much as possible. This is it - especially if people are cloning our toolset in the future.

bebatut:
I agree that duplications should be avoided. But cd-hit was already duplicated before (and before the PRs on IUC) and Pavan needed a fix for this tool, so we fixed it and installed the new TS version. 
I will work on the IUC PRs to make sure cd-hit would be there and then we can "deprecate" the other versions

bgruening:
Since when is a duplicated tool an argument for a tripled version of a tool?
We contacted JJ two month ago and he responded immediately, now we have this tool obviously used in workflows and can not get rid of it anymore.

PR #98, opened 2018-05-30
adds ucsc_fasplit

anilthanki:


tools_iuc.yaml
```diff
@@ -1945,3 +1945,7 @@ tools:
   - name: gtftobed12
     owner: iuc
     tool_panel_section_label: Convert Formats
+ 
+  - name: ucsc_fasplit
+    owner: iuc
+    tool_panel_section_label: Text Manipulation
```
bgruening: There is a similar tools already included: https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/rnateam/splitfasta/rbc_splitfasta/0.2.0

But I guess we should use ucsc tool and deprecate the other.

bgruening:
Its should be installed now.

anilthanki:
Yes they are.. Thanks

PR #113, opened 2018-08-14
Add a first version of ecology tools list

yvanlebras:
- [ ] check if all non-`mnhn65mo` are already installed
  - [ ] it not review and condafy the non-installed once
- [ ] review the `mnhn65mo` based tools
- [ ] check categories

### DIFF
--- a/ecology.yaml
+++ b/ecology.yaml
@@ -4,7 +4,6 @@ install_resolver_dependencies: true
 install_tool_dependencies: false
 
 tools:
-
   - name: 'collapse_collections'
     owner: 'nml'
     tool_panel_section_label: 'Collection Operations'
@@ -56,73 +55,46 @@ tools:
   - name: prinseq
     owner: iuc
     tool_panel_section_label: Text Manipulation
-  - name: add_value
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+  
+ 
   - name: split_file_on_column
     owner: bgruening
     tool_panel_section_label: Text Manipulation
+  
   - name: merge_cols
     owner: devteam
     tool_panel_section_label: Text Manipulation
-  - name: text_processing
-    owner: bgruening
-    tool_panel_section_label: Text Manipulation
-  - name: column_maker
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+  
   - name: coverage
     owner: devteam
     tool_panel_section_label: Text Manipulation
+ 
   - name: regex_replace
     owner: kellrott
     tool_panel_section_label: Text Manipulation
-  - name: tables_arithmetic_operations
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+
   - name: merge
     owner: devteam
     tool_panel_section_label: Text Manipulation
+
   - name: intersect
     owner: devteam
     tool_panel_section_label: Text Manipulation
-  - name: flanking_features
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+
   - name: join
     owner: devteam
     tool_panel_section_label: Text Manipulation
-  - name: cluster
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
-  - name: get_flanks
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
-  - name: complement
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+
   - name: subtract
     owner: devteam
     tool_panel_section_label: Text Manipulation
-  - name: basecoverage
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
+
   - name: subtract_query
     owner: devteam
     tool_panel_section_label: Text Manipulation
-  - name: sed_wrapper
-    owner: bjoern-gruening
-    tool_panel_section_label: Text Manipulation
+
   - name: unique
     owner: bgruening
-    tool_panel_section_label: Text Manipulation
-  - name: dna_filtering
-    owner: devteam
-    tool_panel_section_label: Text Manipulation
-
-
-  - name: column_maker
-    owner: devteam
     tool_panel_section_label: Text Manipulation
 
   - name: reshape2_melt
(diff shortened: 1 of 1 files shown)


ecology.yaml
```diff
@@ -0,0 +1,515 @@
...
+install_resolver_dependencies: true
+install_tool_dependencies: false
+
+tools:
+
+  - name: 'collapse_collections'
+    owner: 'nml'
+    tool_panel_section_label: 'Collection Operations'
+
+  - name: dwt_var_perfeature
```
bgruening: I condafied this one: https://github.com/galaxyproject/tools-devteam/pull/510

ecology.yaml
```diff
@@ -0,0 +1,515 @@
...
+    tool_panel_section_label: Join, Subtract and Group
+
+  # - install_repository_dependencies: true
+  #   install_resolver_dependencies: true
+  #   install_tool_dependencies: false
+  #   name: datamash_wrapper
+  #   owner: agordon
+  #   tool_panel_section_label: Join, Subtract and Group
+
+  - name: suite_datamash
```
bgruening: This one is doubled here and below.

ecology.yaml
```diff
@@ -0,0 +1,515 @@
...
+  - name: collection_column_join
+    owner: iuc
+    tool_panel_section_label: Join, Subtract and Group
+
+  - name: suite_datamash
+    owner: iuc
+    tool_panel_section_label: Join, Subtract and Group
+
+
+  - name: subtract
```
bgruening: This is also multiple times included.

ecology.yaml
```diff
@@ -0,0 +1,515 @@
...
+    tool_panel_section_label: Statistics
+
+  - name: sklearn_suite
+    owner: bgruening
+    tool_panel_section_label: Statistics
+
+
+
+
+  - name: csv_to_tabular
```
bgruening: Should this maybe included in Galaxy as datatype converter. So it can be used implicitly?

ecology.yaml
```diff
@@ -0,0 +1,515 @@
...
+
+  - name: ggplot2_histogram
+    owner: iuc
+    tool_panel_section_label: Graph/Display Data
+
+  - name: tsne
+    owner: iuc
+    tool_panel_section_label: Graph/Display Data
+
+  - name: best_regression_subsets
```
bgruening: We need to check all devteam tools if they are either already installed or if we need to condafy them.

ecology.yaml
yvanlebras: In a way we can use it through "Edit attributes/convert format" history dataset option ? Yes for sure it's a good idea... I never add such a kind of "datatype converter", is there something specific to change on the tool form and/or the Galaxy side ?

ecology.yaml
bgruening: They probably are. If you say all of these are already in. Thats fine. I was just worried by the duplication and one of the first tool is a very old one from devteam that I try to migrate to conda in https://github.com/galaxyproject/tools-devteam/pull/510

ecology.yaml
yvanlebras: Thank you!

ecology.yaml
yvanlebras: eagle eye! Thank you! And sorry for that....

ecology.yaml
yvanlebras: Double eagle eye! Thank you! And sorry again for that....

ecology.yaml
bgruening: Look at https://github.com/galaxyproject/galaxy/tree/dev/lib/galaxy/datatypes/converters and here: https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample#L27

I think these are the only places you need to modify.

ecology.yaml
yvanlebras: ok, thanks!

ecology.yaml
```diff
@@ -0,0 +1,497 @@
...
+
+  - name: biosigner
+    owner: ethevenot
+    tool_panel_section_label: Statistics
+
+  - name: lcmsmatching
+    owner: prog
+    tool_panel_section_label: Annotation
+
+  - name: add_input_name_as_column
```
bgruening: Is this one really needed? Also not updated since 3 years and not condafied.

ecology.yaml
```diff
@@ -0,0 +1,414 @@
...
+
+
+
+
+  - name: csv_to_tabular
+    owner: mnhn65mo
+    tool_panel_section_label: File Conversion
+    
+
+  - name: tabular_to_csv
```
bgruening: Some for this one I guess.
If this is not up-streamed to Galaxy. I guess IUC is a good place to host them.

ecology.yaml
```diff
@@ -0,0 +1,414 @@
...
+    owner: mnhn65mo
+    tool_panel_section_label: File Conversion
+    
+
+  - name: butterfly_analysis
```
bgruening: Do you think the tools from `mnhn65mo` are ready for review? Or do you need more time?

bgruening: I don't think butterfly_analysis is on github?

ecology.yaml
yvanlebras: not needed!

ecology.yaml
yvanlebras: Concerning butterfly_analysis, I think this is the last PR we are waiting from @alanamosse... Tools will be ready for review at the end of the day.

ecology.yaml
yvanlebras: ok I understand. I'm not so familiar with build-in converter, and for sure I need to think more about that, but in a first approach, I think having this as a tool and so the possibility to integrate it on workflows is a good think. So making a PR to IUC will be the best, that's it?

bgruening:
@yvanlebras can you keep this list up to date? Let's aim for the tools that are needed for training ...

yvanlebras:
See up to date list from https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/188

PR #117, opened 2018-08-16
Add Heinz iuc tool

CicoZhang:
Add Heinz iuc tool to UseGalaxy.eu

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -2036,4 +2036,4 @@ tools:
 
   - name: heinz
     owner: iuc
-    tool_panel_section_label: Heinz 
+    tool_panel_section_label: RNA Analysis
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -2036,4 +2036,4 @@ tools:
 
   - name: heinz
     owner: iuc
-    tool_panel_section_label: heinz 
+    tool_panel_section_label: Heinz 
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -2033,3 +2033,7 @@ tools:
   - name: seqtk
     owner: iuc
     tool_panel_section_label: FASTA/FASTQ manipulation
+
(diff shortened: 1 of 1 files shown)


tools_iuc.yaml
```diff
@@ -2036,4 +2036,4 @@ tools:
 
   - name: heinz
     owner: iuc
-    tool_panel_section_label: heinz 
+    tool_panel_section_label: Heinz
```
bgruening: Can we choose on of the existing categories?

bgruening:
Oh, our testing for the categories does not seem to work here. Is this still supposed to work @erasche.

tools_iuc.yaml
CicoZhang: After installing "heinz," 4 tools are installed, which form a workflow (together with the tool 'DESeq2' or 'limma' or something similar); I tend to think that these four tools stay together so that users can access them quickly.

image

After going through all the existing categories in https://usegalaxy.eu/, maybe we can put it into "RNA Analysis," but each of the four tools is not so independent or correlated to RNA (though as a whole, it is).

So maybe it is better to make these 4 tools into one folder (or label) called "Heinz"? (I am flexible with this).

bgruening: I do understand, but we already have so many categories and would like to keep them as small as possible. Also I don't think that the word `Heinz` is giving anyone an impression what that is supposed to do. With `epigenetics` or `transcriptomics` people can at least say, yes this is interesting for me or not.

CicoZhang: No problem, let's put it into ''RNA Analysis". Would be nice if the tool panel can support subfolders (just a naive thought of me) in future. I will change the code now.

bgruening:
Thanks! This will be automatically installed this weekend. Please let us know if you can not see them next week :)

CicoZhang:
No problem! I will keep an eye on that!

PR #130, opened 2018-10-17
check for empty revisions list

ieguinoa:
After running scripts/fix-lockfile.py some tool entries were listed with revisions: [] and were not picked up when using --without option


erasche:
Looks good! Thanks!

PR #131, opened 2018-10-26
Add the norm, csq and cnv tools from the bcftools suite

wm75:


bgruening:
I also installed them in our instance.

PR #133, opened 2018-10-30
Galaxy Proteomics EU Tool YAML Update

jraysajulga:
Added several tools to the YAML file.

### DIFF
--- a/tools_galaxyp.yaml
+++ b/tools_galaxyp.yaml
@@ -39,14 +39,6 @@ tools:
     owner: galaxyp
     tool_panel_section_label: Convert Formats
 
-  - name: peca
-    owner: caleb-easterly
-    tool_panel_section_label: Statistics
-
-  - name: limma_norm
-    owner: caleb-easterly
-    tool_panel_section_label: Statistics
-
   - name: concat_fasta_files
     owner: mandorodriguez
     tool_panel_section_label: FASTA manipulation
@@ -411,10 +403,6 @@ tools:
     owner: jjohnson
     tool_panel_section_label: Beta Test Tools
 
-  - name: metaquant
-    owner: caleb-easterly
-    tool_panel_section_label: Beta Test Tools
-
   - name: protein_transcript_correlation
     owner: pravs
     tool_panel_section_label: Beta Test Tools
@@ -427,16 +415,8 @@ tools:
     owner: galaxyp
     tool_panel_section_label: Beta Test Tools
 
-  - name: proteomic_cravat_score_and_annotate
-    owner: rsajulga
-    tool_panel_section_label: Beta Test Tools
-
   - name: gffutils
     owner: jjohnson
-    tool_panel_section_label: Beta Test Tools
-
-  - name: split_file_to_collection
-    owner: caleb-easterly
     tool_panel_section_label: Beta Test Tools
 
   - name: cravat_score_and_annotate
@@ -661,10 +641,6 @@ tools:
 
   - name: sixgill
     owner: galaxyp
-    tool_panel_section_label: MetaProteomics
-
-  - name: prepare_revigo
-    owner: caleb-easterly
     tool_panel_section_label: MetaProteomics
 
   - name: meta_proteome_analyzer
--- a/tools_galaxyp.yaml
+++ b/tools_galaxyp.yaml
@@ -1,901 +1,161 @@
+---
 install_repository_dependencies: true
 install_resolver_dependencies: true
 install_tool_dependencies: false
 
 tools:
-  - name: retrieve_ensembl_bed
+  - name: uniprotxml_downloader
     owner: galaxyp
     tool_panel_section_label: Get Data
 
-  - name: ms_wiff_loader
+  - name: dbbuilder
     owner: galaxyp
     tool_panel_section_label: Get Data
 
-  - name: regex_find_replace
+  - name: unipept
     owner: galaxyp
-    tool_panel_section_label: Text Manipulation
+    tool_panel_section_label: Get Data
 
-  - name: concatenate_multiple_datasets
-    owner: mvdbeek
-    tool_panel_section_label: Text Manipulation
-
-  - name: filter_by_substring_match
-    owner: alan-blakely
-    tool_panel_section_label: Filter and Sort
-
-  - name: column_join
-    owner: jjohnson
-    tool_panel_section_label: Join, Subtract and Group
-
-  - name: gffcompare_to_bed
+  - name: map_peptides_to_bed
     owner: galaxyp
     tool_panel_section_label: Convert Formats
 
-  - name: bed_to_protein_map
+  - name: idconvert
     owner: galaxyp
     tool_panel_section_label: Convert Formats
 
-  - name: translate_bed_sequences
+  - name: idpquery
     owner: galaxyp
     tool_panel_section_label: Convert Formats
 
-  - name: concat_fasta_files
-    owner: mandorodriguez
-    tool_panel_section_label: FASTA manipulation
-
-  - name: mz_to_sqlite
+  - name: thermo_raw_file_converter
     owner: galaxyp
-    tool_panel_section_label: MS Data Conversion
+    tool_panel_section_label: Convert Formats
 
-  - name: mgf_formatter
+  - name: msconvert_nix
     owner: galaxyp
-    tool_panel_section_label: MS Data Conversion
+    tool_panel_section_label: Convert Formats
 
-  - name: msconvert_win
+  - name: msconvert
     owner: galaxyp
-    tool_panel_section_label: MS Data Conversion
+    tool_panel_section_label: Convert Formats
 
-  - name: ms_data_converter
+  - name: fasta_merge_files_and_filter_unique_sequences
     owner: galaxyp
-    tool_panel_section_label: MS Data Conversion
+    tool_panel_section_label: FASTA/FASTQ manipulation
 
-  - name: proteinpilot
+  - name: peptideshaker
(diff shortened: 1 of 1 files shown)


bgruening:
@jraysajulga this contains a lot more then just proteomics and many tools are already present. But in other files. Can you figure out which tools are really missing in general?

Thanks.

jraysajulga:
@bgruening Sure. I will also do some further organization of these tools. One question: How can we add more sections (e.g., GENERAL TEXT TOOLS, FILE AND META TOOLS) to the tool pane?

bgruening:
@jraysajulga I can point you to this file in a week, currently its in a private repo ... but in general we would really try to avoid new categories as we try to align these with usegalaxy.eu and usegalaxy.org ... we do have `text processing` tools if this is what you need.

jraysajulga:
Okay. That's understandable. We can avoid the new categories then for now.

I just removed tools already present in the tool pane based on this [JSON](https://proteomics.usegalaxy.eu/api/tools?in_panel=true) in that recent commit.

erasche:
@jraysajulga no, that's user controlled. Whatever workflows a user favourites show up there.

bgruening:
@jraysajulga all tools you add to this repo, will be installed on usegalaxy.eu - this is just one server. *.usegalaxy.eu are just different views, where we hide a different set of tools.
So its a little bit more complicated. All IUC tools are in the tools_iuc.yml file and you don't need to add them here. You really only need to add tools that are missing ... Maybe there are two different problems.

1) Are your tools on `usegalaxy.eu` (all tools can be seen there)?
2) If not, should we show them in `proteomics.usegalaxy.eu`

jraysajulga:
@bgruening I removed IUC tools in tools_galaxyp.yaml that were already present in tools_iuc.yaml. However, some IUC tools were not present in the iuc.yaml, so I have some left in the galaxy.yaml. Is that okay? (Example: trinity_gene_to_trans_map)

The majority of tools that I have added are not installed on usegalaxy.eu. We would like to show and use them in proteomics.usegalaxy.eu. Main ones that we want to use for our workflows are as follows:
- regex_find_replace
- custom_pro_db
- gffcompare_to_bed
- translate_bed
- bed_to_protein_map
- concatenate_multiple_datasets
- mz_to_sqlite
- pep_pointer
- peptidegenomiccoordinate

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
-  - name: thermo_raw_file_converter
+  - name: peca
+    owner: caleb-easterly
+    tool_panel_section_label: Statistics
+
+  - name: limma_norm
+    owner: caleb-easterly
+    tool_panel_section_label: Statistics
+
+  - name: concat_fasta_files
```
bgruening: This tool is not maintained and also not needed, it just performs a `cat` operation. This can be done by the normal `cat` tools. The reason is that we do need to maintain the tools over time and for this, we would keep the tool number as much as possible.

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
-    tool_panel_section_label: Proteomics
+    tool_panel_section_label: MS Data Conversion
 
-  - name: lfq_protein_quant
+  - name: validate_fasta_database
     owner: galaxyp
-    tool_panel_section_label: Proteomics
+    tool_panel_section_label: Sequence Database Tools
+
+  - name: fastq_join
```
bgruening: This tool is also not needed imho, it can be replaced by https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/flash/flash/1.2.11.3 I think.

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
+
+  - name: psm_to_sam
+    owner: galaxyp
+    tool_panel_section_label: Proteogenomics
+
+  - name: peptide_to_gff
+    owner: galaxyp
+    tool_panel_section_label: Proteogenomics
+
+  - name: trim_galore
```
bgruening: that should be already installed on usegalaxy.eu

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
     tool_panel_section_label: Convert Formats
 
-  - name: idpquery
+  - name: translate_bed_sequences
     owner: galaxyp
     tool_panel_section_label: Convert Formats
 
-  - name: thermo_raw_file_converter
+  - name: peca
+    owner: caleb-easterly
```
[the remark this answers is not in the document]
jraysajulga: I will work towards getting them (PECA, limma_norm, prepare_revigo)  into the galaxyP repository for now if it's okay with @caleb-easterly and we can add them later.

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
+  - name: directag_and_tagrecon
     owner: galaxyp
-    tool_panel_section_label: Proteomics
+    tool_panel_section_label: Misc
+
+  - name: blosum
+    owner: oscar_b
+    tool_panel_section_label: Beta Test Tools
+
+  - name: select_random_lines
```
bgruening: there is a tool already that does exactly this: https://usegalaxy.eu/root?tool_id=random_lines1

tools_galaxyp.yaml
```diff
@@ -1,158 +1,925 @@
...
+    owner: pravs
+    tool_panel_section_label: Beta Test Tools
+
+  - name: spectrast2spectrast_irt
+    owner: jjohnson
+    tool_panel_section_label: Beta Test Tools
+
+  - name: spectrast
```
bgruening: Is that not also available from galaxyp?

bgruening:
@jraysajulga I would prefer that the IUC tools are going to the IUC file if this is possible, maybe as separate PR? The reason is, that we do update the IUC file automatically and other not that often.

tools_galaxyp.yaml
jraysajulga: I see. Good catch.

bgruening:
@jraysajulga maybe we do this differently.

Lets do this workflow for workflow... you import a workflow and see if tools missing. If these tools are from IUC, devteam or galaxyp no problem we just add them. If they are from different accounts, lets create an issue with the missing tools and we discuss this ... what do you think about this idea?

jraysajulga:
@bgruening  Sounds fair. I recently got a list of workflow tools so this should simplify things quite a bit.

jraysajulga:
Okay. Removed all the fluff and only added only the **galaxyp** tools necessary for our workflows.

There was only one tool in our workflows that was not through galaxyp (don't think it's important enough to raise a GH issue). 

**concatenate_multiple_datasets** by _mvdbeek_

You might have an analogous tool available on EU? Personally, I am not familiar with this tool.

tools_galaxyp.yaml
```diff
@@ -156,3 +159,32 @@ tools:
   - name: proteomics_moff
     owner: galaxyp
     tool_panel_section_label: Proteomics
+
+  - name: regex_find_replace
```
bgruening: @jraysajulga do you know offhand why this tool is stricly needed and can not be replaces by https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_replace_in_line/1.1.1 ... this is my last comment promised :)

jraysajulga: @bgruening concatenate_multiple_datasets can be replaced by another tool so this can be ignored.

As for regex_find_replace, the main issue is that **Replace Text in Entire Line** is unable to take in multiple regex's and replacements. 

In Regex Find and Replace, it is easy to do multiple replacements. We use at least three regular expressions for our workflow. 

Example:
image

tools_galaxyp.yaml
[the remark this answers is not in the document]
bgruening: limma should/could probably go to IUC, there are already limma tools.

Thanks @caleb-easterly!

tools_galaxyp.yaml
bgruening: @jraysajulga can you not simply run 3 tools of the classic regex-replace? Or do I miss something?

jraysajulga: @bgruening We could do that if needed.

Is that okay,  @subinamehta?

subinamehta: We could that, but wouldn't the tool be more useful if people have multiple conditions to be fulfilled?

bgruening: @subinamehta yes. But we should rather update the old tool and this is not a critical missing feature or is it. I don't have hard feelings about this. Sorry, I just try to avoid many tools with the same functionality in Galaxy. I would rather try to maintain as less as possible tools over the years to come :)

jraysajulga: @subinamehta We can raise that as a requested functionality in the **Replace Text in entire line** tool. Would that be a GitHub issue?

@bgruening Understandable. I have removed the regex_find_replace tool from yams file.

bgruening:
Thanks! :)

bgruening:
@subinamehta If I'm not mistaken its this tool: https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/text_processing/find_and_replace.xml any contribution welcome.

jraysajulga:
Thanks for installing the tools, @bgruening. Most of them look good except for one: **custom_pro_db**. Can you check this?

bgruening:
@jraysajulga did I forget anything here? Is custom_pro_db still a problem?

jraysajulga:
@bgruening Thanks for checking up... Looks good now!

PR #141, opened 2018-12-03
Add missing tools for RNA workbench

bebatut:
ping @bgruening, @eggzilla, @jfallmann

### DIFF
--- a/rnateam.yml
+++ b/rnateam.yml
@@ -27,3 +27,47 @@ tools:
   - name: sshmm
     owner: rnateam
     tool_panel_section_label: RNA Analysis
+
+  - name: cmv
+    owner: rnateam
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: rnalien
+    owner: rnateam
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: fuma
+    owner: yhoogstrate
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: infernal
+    owner: bgruening
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: rnammer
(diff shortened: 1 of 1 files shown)


rnateam.yml
```diff
@@ -27,3 +27,47 @@ tools:
...
+  - name: fuma
+    owner: yhoogstrate
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: infernal
+    owner: bgruening
+    tool_panel_section_label: RNA Analysis 
+    
+  - name: rnammer
+    owner: lionelguy
```
bgruening: is that true?

jfallmann: looks good, but there will be more to add, hopefully soon

bebatut: Same as in workbench: https://github.com/bgruening/galaxy-rna-workbench/blob/0de0d004bc3fc9a22761ed5713abc0c2a480ec81/rna_workbench_2.yml#L85

bgruening: It was updated last 2013, we should take it over or let it die ...

bgruening: Sorry, I'm more strict on EU than on our Docker.

jfallmann: is there a reason why we wouldn't take it over? Sorry maybe I missed the point

bgruening: No reason, just time. Make sure a conda-package is there and tests and get it into galaxytools.

bgruening: I removed it for now, please feel free to take it over to galaxytools.

bgruening:
Installation is triggered manually. We will have all these tools soon on https://rna.usegalaxy.eu

PR #147, opened 2019-01-08
Add velvet from toolshed

abretaud:
There are already some velvet tools in usegalaxy.eu, but they are not up-to-date with the ones in the toolshed. As they are used for the [assembly introduction in GTN](https://galaxyproject.github.io/training-material/topics/assembly/), currently only genouest and bipaa are listed as valid galaxy servers for this tutorial. It would be good to have at least one *big* instance listed I guess

### DIFF
--- a/training.yaml
+++ b/training.yaml
@@ -7,3 +7,6 @@ tools:
   - name: velvetoptimiser
     owner: simon-gladman
     tool_panel_section_label: Assembly
+  - name: velvet
+    owner: devteam
+    tool_panel_section_label: Assembly


bgruening:
I'm always confused about what is the latest version :(
Should we change maybe the training to use the build-in wrappers? I'm a little bit hesitant to install a tool that really should not be used anymore :) I labeled the included tools already as `training-only` :)

@Slugger70 can you comment here?

Thanks, @abretaud!

Slugger70:
Velvet and VelvetOptimiser should definitely not be used for real work at all anymore. The reason it is in the tutorial is so that users can see what happens when you build a de Bruijn graph with just one k-mer size. The compare the outcomes with different k-mer sizes a bit later on. Velvet is just so convenient for this. Spades has too many other things going on and so is not that good for illustrating how de Bruijn graphs work. So it's a bit of a conundrum.

I still get users of GalaxyAustralia using VelvetOptimiser and Velvet for real work no matter how many times I tell them not to so I'm considering removing it off my server too! Maybe we should put some validation into the tool wrapper that looks at the input file and checks if it's a training dataset and refuses to run otherwise or something.

bgruening:
@Slugger70 I'm fine with keeping one tool. We have added a huge scary label - that is ok I think.
Please go to usegalaxy.eu and search for velvet. What I'm not sure is the installed tools on usegalaxy.eu are from the Galaxy source and not the TS. The added tool in this PR is from devteam, so which tool is really needed and do we need to install the devteam tool as well? And remove the local once? I'm only interested to make the training work - but I'm confused which one is the correct one.

erasche:
Things worked fine last time preparing for training with the velvets + VO that were installed, so closing this unless there's some special reason we need that one

PR #151, opened 2019-01-26
Delete update_trusted_tools.sh

bgruening:
I'm not sure if this is used somewhere else, but I think this is deprecated and can be removed.


erasche:
No, definitely unused :)

PR #153, opened 2019-01-29
added tools for RNA analysis found only in the RNA workbench

bagnacan:
Added tools that are found on the Galaxy RNA workbench, but missing from rna.usegalaxy.eu:
- Dr. Disco
- CMCompare
- MEA
- RNAammer

### DIFF
--- a/rnateam.yaml
+++ b/rnateam.yaml
@@ -76,7 +76,6 @@ tools:
     owner: rnateam
     tool_panel_section_label: RNA Analysis
 
-
   - name: cmcv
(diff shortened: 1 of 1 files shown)


rnateam.yaml
```diff
@@ -20,6 +20,10 @@ tools:
     owner: rnateam
     tool_panel_section_label: Annotation
 
+  - name: rnammer
```
bgruening: I'm not sure this is maintained. We should not include this, it's not from us.

PR #155, opened 2019-01-31
Add mea and cofold for the RNA workbench

bebatut:
ping @pavanvidem


bebatut:
I put my changes in #153

PR #157, opened 2019-02-09
add first batch of stable imaging tools

bgruening:
ping @ThomasWollmann


bgruening:
This will be installed next weekend if I do not trigger it earlier.

Installation kind of depends on https://github.com/usegalaxy-eu/infrastructure-playbook/pull/33

PR #160, opened 2019-03-04
Add scanpy

bebatut:
If https://github.com/usegalaxy-eu/infrastructure-playbook/pull/36 is ok and once scanpy is on the toolshed.

Should we add also a `beta` label to it? Is it possible?


bgruening:
@bebatut we can do this, but currently only manually. Can you please add all tools separately? The suite stuff is in my experience not really working well.

bgruening:
@erasche without adding the category to the schema this PR should have been failed isn't it?

bgruening:
Yes, exactly. So known issue?

Thanks @bebatut!

PR #176, opened 2019-04-24
Things needed for genome annotation

erasche:
Ping @abretaud 

Just moved augustus from europe-custom to annotation since it's a better fit there.


abretaud:
Thanks @erasche!

PR #177, opened 2019-04-29
Update tools_galaxyp.yaml

subinamehta:
added metaquantome suite to the tools list


bgruening:
@subinamehta can you please add all the single tools separately as well. Thanks!

PR #180, opened 2019-05-13
Clean up fastqc versions

wm75:
Remove outdated fastqc wrapper revisions with tool wrapper version 0.72.
Revision `9da02be9c6cc` represents the approved 0.72 version.

### DIFF
--- a/tools_iuc.yaml.lock
+++ b/tools_iuc.yaml.lock
@@ -989,9 +989,7 @@ tools:
 - name: fastqc
   owner: devteam
   revisions:
-  - 3e1cdf5406db
   - 484e86282f4b
-  - c15237684a01
   - e28c965eeed4
   - 9da02be9c6cc
   tool_panel_section_label: FASTA/FASTQ manipulation


wm75:
Revision `3e1cdf5406db` featured some tool wrapper enhancements that got rolled back, but will be recommitted under a new tool wrapper version 0.72+galaxy1 soon. (compare https://github.com/galaxyproject/tools-iuc/pull/2406 and https://github.com/galaxyproject/tools-iuc/pull/2410).

wm75:
Hope this is the right place to fix this version mess?

erasche:
This is part of the fix, yes. Should we uninstall those versions as well? Or hide them somehow?

wm75:
I would vote for uninstalling them cause that would prevent confusion in the future. I'm not sure whether the new `9da02be9c6cc` is actually installed already (it's not accessible at least). If not that needs to be done, too.

wm75:
Thanks for the fast response :)

erasche:
Removed, added a function to bioblend to help this be faster in future https://github.com/galaxyproject/ephemeris/issues/83 fyi @bgruening, maybe if we get marius to implement it in ephemeris we can have a 'remove.yml' where we put revisions we don't want and have them auto-removed

PR #181, opened 2019-05-13
Update augustus_training

abretaud:
Ok, this one did not get auto updated as it's not owned by iuc.
For snap_training, the lock file was autoupdated, but the new version is not on usegalaxy.eu, is it normal?

### DIFF
--- a/genome-annotation.yaml.lock
+++ b/genome-annotation.yaml.lock
@@ -26,6 +26,7 @@ tools:
   owner: bgruening
   revisions:
   - 86c89c3bd99d
+  - 1fbb1135da16
   tool_panel_section_label: Annotation
 - name: fasta_stats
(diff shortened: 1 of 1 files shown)


erasche:
Not normal, will look into.

erasche:
```
(13/13) Installing repository snap_training from iuc to section "Annotation" at revision 2a598c0aa042 (TRT: 0:01:12.217716)
	* Error installing a repository (after 0:00:08.706263 seconds)! Name: snap_training,owner: iuc, revision: 2a598c0aa042, error: {"err_msg": "Uncaught exception in exposed API method:", "err_code": 0}
```

of course

erasche:
Fixed the issue internally, re-running now, looks good!

PR #194, opened 2019-06-20
new ecology tools

bgruening:
ping @yvanlebras 

We need a category for it ... `Ecology`?


yvanlebras:
hmmmm...
Ecology can be ok as this is really dedicated to the ecology field.. Here it's abundance index calculation... 
Maybe `Species abundance` ?

yvanlebras:
if we want to give some precisions....

bgruening:
Your call ... keep in mind to not create too many sections.

Let me know and I will adopt it.

yvanlebras:
I think `Species abundance` is a good section name where we will put a lot of tools.
Thank you

PR #197, opened 2019-06-21
sort the values

bgruening:
@erasche I can not test this as I have shitty internet, but maybe its good as it is. It will hopefully prevent the commits that do not change anything just the ordering.


erasche:
oh, good, that should have been there.

PR #198, opened 2019-06-26
Create single-cell-ebi-gxa.yaml for Human Cell Atlas flavour

pcm32:
Here are the tools that we developed through funding by the Human Cell Atlas, in the framework of the Single Cell Portals (CZI).

### DIFF
--- a/single-cell-ebi-gxa.yaml
+++ b/single-cell-ebi-gxa.yaml
@@ -7,191 +7,151 @@ tools:
   - name: dropletutils_read_10x
     owner: ebi-gxa
     tool_panel_section_label: 'HCA-Single Cell'
-    tool_shed_url: https://toolshed.g2.bx.psu.edu/
     
   - name: gtf2gene_list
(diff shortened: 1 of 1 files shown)


single-cell-ebi-gxa.yaml
```diff
@@ -0,0 +1,196 @@
+---
+install_repository_dependencies: true
+install_resolver_dependencies: true
+install_tool_dependencies: false
+
+tools:
+  - name: dropletutils_read_10x
+    owner: ebi-gxa
+    tool_panel_section_label: 'Single Cell'
```
bgruening: Can we make this "Human Cell Atlas" or something different? We do have a category currently called single-cell and we can not filter based on tools-ids (just now). So we filter per sections atm.

pcm32: Would 'HCA-Single Cell' do?

bgruening: Think so yes.

PR #199, opened 2019-06-27
move RaceID tools under SingleCell

mtekman:


bgruening:
Ok adopted it also in the shed_tool_conf

PR #202, opened 2019-07-01
Update epigenetics.yaml

kpoterlowicz:
Tools need to run ewas training

### DIFF
--- a/epigenetics.yaml
+++ b/epigenetics.yaml
@@ -34,11 +34,7 @@ tools:
   - name: suite_deeptools
     owner: bgruening
     tool_panel_section_label: "Epigenetics"
-    
-  - name: ewastools
-    owner: kpbioteam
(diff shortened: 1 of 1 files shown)


epigenetics.yaml
```diff
@@ -34,3 +34,20 @@ tools:
   - name: suite_deeptools
     owner: bgruening
     tool_panel_section_label: "Epigenetics"
+    
+  - name: ewastools
```
bgruening: please add all tools separately ...

kpoterlowicz: Ok added

erasche:
@kpoterlowicz @bgruening none of these are on the main toolshed. they have not been installed. Nothing in this repo can be installed from TTS, it's all hardcoded to MTS.

I will comment these out for now.

PR #205, opened 2019-07-04
Add Stacks2 tool suite

yvanlebras:


bgruening:
Muha, this is now working right? @yvanlebras @bernt-matthias do you want to write a small paragraph about those tools at https://github.com/usegalaxy-eu/website/tree/master/_posts it will be linked to the Galaxy Hub as well.

yvanlebras:
True, it works wonderfully! Thank you so much Matthias! I think this is a very good idea Björn. You can start something Matthias ?

bernt-matthias:
Will think about it. Are there any changes needed in the gtn tutorial or the work flow shipped with the tutorial? Then we would have the complete package..

bgruening:
That is a question for @yvanlebras. Imho he tried the tutorial and it worked fine for him.

yvanlebras:
with stacks2 the training was ok. I have to update the material with this new version

bernt-matthias:
I think training update should be done first.

PR #215, opened 2019-08-01
Added unzip tool

ThomasWollmann:


ThomasWollmann:
@bgruening I think this tool is very useful for many use cases. Not just imaging.

PR #224, opened 2019-09-19
add datamesh for Climate "instance"

annefou:
We are working on Galaxy 101 for everyone where datamash is used so it would be nice to have these tools for the climate "instance" too.
Thanks.


bgruening:
Thanks!

PR #235, opened 2019-10-16
Update tools_galaxyp.yaml

subinamehta:
adding a new tool ..maxbin2
https://toolshed.g2.bx.psu.edu/repository/view_repository?changeset_revision=864279a0d64b&id=7124d41fcda91a0c

### DIFF
--- a/tools_galaxyp.yaml
+++ b/tools_galaxyp.yaml
@@ -210,4 +210,8 @@ tools:
     
   - name: metaquantome_viz
     owner: galaxyp
-    tool_panel_section_label: Proteomics  
+    tool_panel_section_label: Proteomics 
+    
+  - name: maxbin2
+    owner: mbernt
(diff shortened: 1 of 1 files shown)


tools_galaxyp.yaml
```diff
@@ -210,4 +210,8 @@ tools:
     
   - name: metaquantome_viz
     owner: galaxyp
-    tool_panel_section_label: Proteomics  
+    tool_panel_section_label: Proteomics 
+    
+  - name: maxbin2
+    owner: mbernt
```
bgruening: @mbernt can we move this to IUC or GalaxyP?

We would need to track yet another repo for creating containers and so on ...

bgruening: This is now in IUC ... thanks @bernt-matthias

PR #238, opened 2019-10-23
Add missing tools from samtools suite

mvdbeek:
Many of these are tools added by @bernt-matthias and are very useful.

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -1434,6 +1434,10 @@ tools:
     owner: iuc
     tool_panel_section_label: SAM Tools
 
+  - name: samtools_view
+    owner: iuc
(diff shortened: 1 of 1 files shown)


bgruening:
Gracias!

bernt-matthias:
@pvanheus was missing `samtools view` on usegalaxy.eu (https://github.com/galaxyproject/tools-iuc/pull/2649)

I also would suggest to deprecate `Filter SAM or BAM, output SAM or BAM files on FLAG MAPQ RG LN or by region` here and at IUC in favor of samtools view.

bgruening:
We still have some problems with installing tools in 19.09, sorry, this will deployed sometime soon. Or I just do it manually for now.

bgruening:
Should be online now ...

PR #243, opened 2019-11-01
add train test split and other 2 tools

qiagu:


simonbray:
@bgruening I think these were not added, right?

bgruening:
@simonbray what do you mean? Are they not installed?

PR #245, opened 2019-11-07
Update europe-custom.yaml

gmauro:
Add exomedepth


europe-custom.yaml
```diff
@@ -1171,6 +1171,9 @@ tools:
   - name: vcftools_subset
     owner: devteam
     tool_panel_section_label: "Variant Calling"
+  - name: exomedepth
```
bgruening: This tool has no tests and does not adhere to the IUC best practices ... can we move it to IUC before we install it?

gmauro:
Nothing to change here, isn't it?

PR #246, opened 2019-11-10
add snp_sites and snp_dists tools

pvanheus:
* snp-sites works with multiple sequence alignments and can also extract a count of constant sites (see https://bitsandbugs.org/2019/11/06/two-easy-ways-to-run-iq-tree-with-the-correct-number-of-constant-sites/)
* snp-dists computes a SNP distance matrix from a multiple sequence alignment


bgruening:
Merci!

PR #252, opened 2019-12-06
HCA additional tools

pcm32:
This PR adds SCCAF multiple regress out, seurat 3 find neighbours, and ann_data_ops for metadata modifications of AnnData objects.


pcm32:
I think that this is obsolete given #264

PR #262, opened 2020-01-07
Update augustus

abretaud:
Updating to latest augustus version (not done by jenkins as it's owned by bgruening) from https://github.com/galaxyproject/tools-iuc/pull/2759

### DIFF
--- a/genome-annotation.yaml.lock
+++ b/genome-annotation.yaml.lock
@@ -64,12 +64,14 @@ tools:
   owner: bgruening
   revisions:
   - f5075dee9d6b
+  - ca6d970d931c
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks! The problem is not the stupid owner bgruening but that this file is missing here: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L35

I guess it is save to add genome-annotation as trusted repo. I will do this. Thanks for the update.

abretaud:
Ah ok, it makes sense!
I'll test it on pulsar when it's updated on usegalaxy.eu

PR #272, opened 2020-02-01
Update psymap

annefou:
upgrade pysmap climate tool to the lastest version.


climate.yaml.lock
```diff
@@ -5,7 +5,7 @@ tools:
 - name: psy_maps
   owner: climate
   revisions:
-  - db8d76da4174
+  - 706666d912d5
```
bgruening: Please leave the other version in.. just add new ones.

annefou:
@bgruening  Ok. Sorry I did not know. So I put the previous version back. Are we supposed to add newer version after or before? I add it after. 
Thanks!

bgruening:
> @bgruening Ok. Sorry I did not know. So I put the previous version back. Are we supposed to add newer version after or before? I add it after.
> Thanks!

The ordering does not matter. If you say that all your tool updates are following the IUC best practice
guideline :) than we can auto-update your tools :)

PR #285, opened 2020-03-23
Update cheminformatics.yaml

bgruening:
@simonbray it seems we are missing a few of our own tools :(


simonbray:
Okay, thanks for the update

PR #289, opened 2020-04-10
Create proteore_tools.yaml

jj-umn:
Add ProteoRE tools for https://proteomics.usegalaxy.eu

### DIFF
--- a/proteore_tools.yaml
+++ b/proteore_tools.yaml
@@ -1,0 +1,92 @@
+---
+install_repository_dependencies: true
+install_resolver_dependencies: true
+install_tool_dependencies: false
+
+tools:
+
+## ProteoRE Data Manipulation
+
+  - name: proteore_id_converter
+    owner: proteore
+    tool_panel_section_label: ProteoRE
+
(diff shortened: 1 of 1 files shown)


bgruening:
@jj-umn we can not add arbitrary tools to EU. Do those tools follow the IUC guidelines? Why are they not in GalaxyP? I assume we also do not generate Containers for those tools?

jj-umn:
@bgruening @yvandenb @dchristiany   This is an initial request to add ProteoRE tools to https://proteomics.usegalaxy.eu  from https://github.com/ifb-git/ProteoRE
The ProteoRE team can change this as fits their needs.  I don't know that it is possible to insert labels within sections from this proteore_tools.yaml file.  My experience is that label to be added manually to the config/tool_conf.xml and the generated integrated_tool_panel.xml  
@yvandenb I'll let you take this up from here.

jj-umn:
@bgruening I'm just trying to help the IFB ProteoRE group get started on this.  They can take it from here.

PR #298, opened 2020-05-18
Remove SEACR

stephanflemming:
Wrong tool_id. Duplicate entry of SEACR removed.


bgruening:
Mh, this is actually the latest ID ... there was one older ... Ok, maybe this PR is not needed.

PR #299, opened 2020-05-22
Add xarray tools to climate Galaxy

annefou:
I am not sure about xarray_select
Even though it is in https://github.com/galaxyecology/tools-ecology/tree/master/tools/data_manipulation/xarray I can't find it in the toolshed.


bgruening:
@annefou can you help me here: https://github.com/galaxyecology/tools-ecology/pull/23 ?

PR #303, opened 2020-05-25
Add xarray tools to ecology flavour

yvanlebras:


bgruening:
I know this is confusing, we will use this tool list in the future for the sub-domains, but currently, you need to edit it here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/templates/galaxy/config/global_host_filters.py.j2

Which might not be possible, as you need to "show" the climate category :(

yvanlebras:
Thank you Björn, I was researching the link of this file... Thanks !
ITM, it appears that xarray tools are on the usegalaxy.eu "climate_analysis" tool section, I think this can be more appropriate to put these xarray tools on the "GIS data handling" one (or maybe in both ?).

bgruening:
both is at the moment not possible ... @annefou is the GIS section ok for you as well?

yvanlebras:
I can understand that both is not possible for now. 
NetCDF is a data format, so not so easy to classify such related tools on 1 particular tool section. But, this format is used particularly in climate, meteorology, oceanography... in a way that can be IMO summarized as GIS... Waiting for the feedback from Anne.
THANK you again Björn, Thank you Anne !!!

annefou:
GIS section is OK for me.

bgruening:
@yvanlebras can you change the section in the climate yml please as well?

yvanlebras:
Done! Thank you

bgruening:
Ok, tomorrow this should be fixed.

PR #313, opened 2020-07-27
Move medaka tools to Nanopore

willemdek11:
This way, they'll also be available at nanopore.usegalaxy.eu


bgruening:
Thanks @willemdek11!

PR #315, opened 2020-08-07
Add apoc and smina (from Earlham Institute)

simonbray:
@nsoranzo, I hope you have no objection if we install these tools on usegalaxy.eu?


nsoranzo:
Of course, more than welcome!

PR #317, opened 2020-08-22
Update cheminformatics.yaml

simonbray:


simonbray:
@bgruening, sorry for the inconvenience, but would it be possible to manually install these tools before Saturday?

bgruening:
Yes.

PR #322, opened 2020-09-05
Add new legacy_tools.yaml

wm75:
This is intended for outdated, hidden tools that should not receive automated updates anymore.

Migrates sam-to-bam and bam-to-sam as the first tools into the new file since these two are very old tools that can be replaced fully with samtools_view.

Hope this approach is correct?


bgruening:
Yes, but you also need to cover the lock file as well.

wm75:
Ah yes, will do.

PR #324, opened 2020-09-06
Keep extract_genomic_dna in Fetch Sequences

wm75:


europe-custom.yaml
```diff
@@ -242,7 +242,7 @@ tools:
     tool_panel_section_label: Extract Features
   - name: extract_genomic_dna
     owner: iuc
-    tool_panel_section_label: "Other Tools"
+    tool_panel_section_label: "Fetch Sequences / Alignments"
```
bgruening, line 245: With special characters it might be better to specify not the `tool_panel_section_label` but the `tool_panel_section_id`

bgruening: Reason is that Galaxy has some magic to convert those labels to IDs, but with special characters, you never know ...

wm75: I see. With `tool_panel_section_id` as the yaml key, then?

bgruening: Yes: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/single-cell-ebi-gxa.yaml.lock#L12

bgruening (approved):

bgruening:
Danke.

PR #348, opened 2020-11-03
Add CI for linting

simonbray:
See the failing PRs on my fork which show how this works: https://github.com/simonbray/usegalaxy-eu-tools/pulls

I am just keeping it simple and testing that altered yaml files are openable and added tools are in the toolshed. I experimented with yamllint as well but this returns a very large number of errors and I'm not sure we need to care about these.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+name: Tool linting
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Perform linting
+        run: make pr_check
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Perform linting
+        run: make pr_check
```
bgruening, line 28: do you think we should also run `make lint`, it is supposed to check that the categories are correct.

bgruening (commented):
Thanks @simonbray!

.github/workflows/lint.yml
simonbray: It looks like `make lint` is just updating the lock files? I'm not sure this should be run in the PR.

bgruening: Its also checking against a schema.yml file if all categories are actually defined.

Makefile
```diff
@@ -31,6 +31,8 @@ install: $(INSTALL_YAMLS) ## Install the tools in our galaxy
 	@echo "Installing any updated versions of $<"
 	@-shed-tools install --install_resolver_dependencies --toolsfile $< --galaxy $(GALAXY_SERVER) --api_key $(GALAXY_API_KEY) 2>&1 | tee -a report.log
 
+pr_check:
+	for changed_yaml in `git diff remotes/origin/master --name-only | grep .yaml$$`; do python scripts/pr-check.py $${changed_yaml}; done
```
simonbray, line 35:
```suggestion
	for changed_yaml in `git diff remotes/origin/master --name-only | grep .yaml$$`; do python scripts/pr-check.py $${changed_yaml} && pykwalify -d $${changed_yaml} -s .schema.yaml ; done
```

simonbray: How about this @bgruening?

bgruening:
Ok, cool let's test this! Thanks a lot @simonbray!

PR #352, opened 2020-11-18
Update SISTR on Galaxy-EU to v1.1.1

kbessonov1984:
Please update SISTR to the newest version 1.1.1 (current one is 1.0.2 on the server). Thank you for having our tool on your sever.


bgruening:
@kbessonov1984 thanks! This way you can also add new tools. We will install tools automatically every Saturday. If you need them earlier, let me know.

PR #358, opened 2020-11-30
Included tool GECKO

estebanpw:
This PR is to include the tool GECKO, which is already available in bioconda and tools-iuc. 
It has been included right below the CHROMEISTER tool, under category Multiple Alignments


bgruening:
Cool, thanks.

PR #365, opened 2021-01-15
Add mztab2tsv tool

annalefarova:
Add mztab2tsv tool into the European Galaxy server.


bgruening.yaml
```diff
@@ -195,3 +195,7 @@ tools:
   - name: add_line_to_file
     owner: bgruening
     tool_panel_section_label: Text Manipulation
+
+  - name: mztab2tsv
+    owner: bgruening
```
bgruening, line 200: you can add this to the galaxyp file with section Proteomics I think or `Convert Formats`

PR #368, opened 2021-02-04
remove gmx_npt and gmx_nvt

gmauro:
remove gmx_npt and gmx_nvt as discussed with @simonbray 
These tools run using all cores available in the computing node, making it unusable.

### DIFF
--- a/cheminformatics.yaml
+++ b/cheminformatics.yaml
@@ -129,10 +129,6 @@ tools:
     owner: bgruening
     tool_panel_section_label: ChemicalToolBox
   
-  - name: gmx_npt
-    owner: chemteam
-    tool_panel_section_label: ChemicalToolBox
-
   - name: md_converter
     owner: chemteam
     tool_panel_section_label: ChemicalToolBox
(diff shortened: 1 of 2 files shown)


bgruening:
This will not remove the tools. Removing is not implemented at the moment. Is here a way how we can fix those tools?

Is hiding better?

simonbray:
We could add a set number of cores to the tools here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/dynamic_rules/usegalaxy/tool_destinations.yaml. That would 'fix' the problem.

However, no-one should be using these tools, so it would be better to hide them IMO.

gmauro:
Remove "from the tools' menu"

simonbray:
> Remove "from the tools' menu"

Or block tool execution. I believe someone said this was an option.

bgruening:
Can you see if the tools are.gone?

simonbray:
They are not findable via search, but you can still navigate directly to e.g. https://usegalaxy.eu/?tool_id=gmx_nvt and the tool can also be executed.

bgruening:
Yes they are only hidden. If you really want to remove it we can remove it. I will do this tonoght.

simonbray:
> Yes they are only hidden. If you really want to remove it we can remove it. I will do this tonoght.

I thought this was discouraged in case they are used in workflows?

If it's possible, yes, it would be good if they can be removed - in that case can you also remove `gmx_md` as well please?

bgruening:
Removed from the server. It probably needs to resync.

PR #379, opened 2021-04-20
Add metaphlan tool suite

bebatut:
Should I add the data manager there too?

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -3391,3 +3391,19 @@ tools:
   - name: pe_histogram
     owner: iuc
     tool_panel_section_label: Graph/Display Data
+  
+  - name: metaphlan
+    owner: iuc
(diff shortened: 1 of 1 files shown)


bgruening:
Yes please add data manager as well and remove the older one, also from the .lock file please.

tools_iuc.yaml
```diff
@@ -3391,3 +3391,26 @@ tools:
   - name: pe_histogram
     owner: iuc
     tool_panel_section_label: Graph/Display Data
+  
+  - name: metaphlan
```
bgruening, line 3395: I have 12 new repos for human related tools ...

bebatut: I wanted to add with a new PR :)

bgruening: a 14 ...

PR #380, opened 2021-04-20
fix complement

gmauro:
remove a buggy version and duplication

### DIFF
--- a/tools_iuc.yaml.lock
+++ b/tools_iuc.yaml.lock
@@ -1144,7 +1144,6 @@ tools:
   revisions:
   - 38c8bb402872
   - d958d5a0d1e8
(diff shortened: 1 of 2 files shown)


tools_iuc.yaml.lock
```diff
@@ -1144,7 +1144,6 @@ tools:
   revisions:
   - 38c8bb402872
   - d958d5a0d1e8
-  - e0a23ab32d7f
```
bgruening, line 1147: @gmauro you need to manually remove this tool as ephemeris will not deinstall tools atm.

gmauro: yes, I have done it

PR #383, opened 2021-05-12
Update humann tool suite

bebatut:


tools_iuc.yaml
```diff
@@ -3364,62 +3364,65 @@ tools:
     owner: iuc
     tool_panel_section_label: RAD-seq
 
-  - name: humann2
+  - name: humann
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
 
-  - name: humann2_associate
+  - name: humann_associate
```
bgruening, line 3371: I would leave the old tools, we don't want to remove them left? We want to hide them?

Side note: removing tools from the yaml does not remove it from Galaxy. This is not implemented.

bebatut: ok. I will revert that then

bebatut:
I do not understand why it is failing. `humann`  is on the MTS (https://toolshed.g2.bx.psu.edu/view/iuc/humann/65c80ca30373)

PR #387, opened 2021-06-04
Add dfornika's ivar_variants_to_vcf

pvanheus:
This is useful to feed output of ivar variants to SnpEff. The script is originally from the [viralrecon](https://github.com/nf-core/viralrecon) project.

### DIFF
--- a/diverse.yaml
+++ b/diverse.yaml
@@ -10,4 +10,5 @@ tools:
 
   - name:  ivar_variants_to_vcf
     owner: dfornika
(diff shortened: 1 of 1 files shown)


pvanheus:
This is failing lint - is only main toolshed supported?

bgruening:
@pvanheus we don't install tools from the test toolshed, sorry. We need to guarantee a certain level of quality and don't want to break tool lineage if possible.

bgruening:
@dfornika do you need help with this tool to get it into the TS? Or do you want to submit this to IUC?

PR #389, opened 2021-06-08
update glimmer

gmauro:
This commit removes an old/custom version of glimmer and install the latest suite available in the IUC repo

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -1635,6 +1635,10 @@ tools:
     owner: devteam
     tool_panel_section_label: Phenotype Association
 
+  - name:  suite_glimmer
+    owner: bgruening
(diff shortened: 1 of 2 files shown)


tools_iuc.yaml
```diff
@@ -1635,6 +1635,10 @@ tools:
     owner: devteam
     tool_panel_section_label: Phenotype Association
 
+  - name:  suite_glimmer
```
bgruening, line 1638: Pleas do not install suite but the tools individual y. Suite don't work really.

gmauro: I was in doubt indeed

bgruening (approved):

bgruening:
@gmauro you need to manually remove this old version and install it manually if you need it before Saturday.

PR #390, opened 2021-06-13
update fastqc in the FASTQ Quality Control section'

gmauro:
This PR will add the last fastqc version in the FASTQ Quality Control section but, I am puzzled by the fact we have this section with only one tool ,and it is completely different from others usegalaxy.* server.


bgruening:
FASTQ Quality Control should go away imho. We should move everything into the FASTA/FASTQ section.

PR #391, opened 2021-06-15
Add CESM tool to Galaxy climate

annefou:
This tool is multi-processors (Message Passing Interface). This is our first "parallel" tool so I am not sure what we need to do to be able to use it properly in Galaxy Europe.


bgruening:
Is installed in 10min

annefou:
Cool! How do we know how much resources is dedicated to the tool? And how much can you give to a single tool? 
Thank you!!!

bgruening:
I don't know ;) You need to tell me. The upper limit is something about 120CPU and 2TB memory per tool. But I would not recommend that ;)

annefou:
Is 64 CPUs & 128GB reasonable?

If we have too little, it is scientifically not relevant. But we could also make tests first with much smaller number of processors.

bgruening:
64? what is a lot :)

You can have this in theory, the problem with that is that you can not run many tools in parallel then. Can you start with something low. 12, 24 and see how this performer?

You can configure this here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/dynamic_rules/usegalaxy/tool_destinations.yaml

gmauro:
I am not sure that you are speaking the same tongue here.
MPI is completely different from Multicores.

gmauro:
> I don't know ;) You need to tell me. The upper limit is something about 120CPU and 2TB memory per tool. But I would not recommend that ;)

No, that's not true! You can't use 'and' but 'or'

annefou:
Ok. I don't fully understand. I see tools where OMP_NUM_THREADS is used so I thought I could use MPI in a similar way. This is what we do on Virtual machine. Is it because HT condor does not support MPI?

gmauro:
HTCondor yes, the condor runner not (FWIK)
The condor runner is using the 'vanilla' universe and not the 'parallel'

gmauro:
but seems there is a way to run parallel applications with the vanilla universe also
https://htcondor.readthedocs.io/en/latest/users-manual/parallel-applications.html?highlight=mpi#mpi-applications-within-htcondor-s-vanilla-universe

gmauro:
Do you want to run a job that spans multiple machines, where the multiple processes running concurrently on these multiple machines, communicating with each other, right?

annefou:
Yes I read this HTcondor stuff and I don't really understand it yet...
We usually span on multiple machines (this is how we run on an HPC) but I suspected it would not be possible with Galaxy so I configured the tool to run on one single node but multi-processors on this single node/machine. So I used `GALAXY_SLOTS` to tell my tool that it can use `GALAXY_SLOTS` processors per node. This is also how we run this tool on virtual machines.

gmauro:
Fine, now it's clear.

PR #399, opened 2021-07-21
added new tool: imaging > spot_detection_2d

qgao-hd:
imaging.yaml has been edited.

### DIFF
--- a/imaging.yaml
+++ b/imaging.yaml
@@ -47,6 +47,10 @@ tools:
     owner: imgteam
     tool_panel_section_label: "Imaging"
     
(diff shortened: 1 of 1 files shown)


bgruening:
And now its passed :) 

Thanks!

bgruening:
It will be installed on Saturday if you don't need it earlier.

qgao-hd:
> It will be installed on Saturday if you don't need it earlier.

Thanks. It will be great if this and the following two tools could all be installed on Saturday. Then I can start making the new workflows for our partners.

bgruening:
They should be installed.

PR #406, opened 2021-08-13
Add Data exploration tools

yvanlebras:


ecology.yaml
```diff
@@ -71,6 +71,30 @@ tools:
     owner: ecology
     tool_panel_section_label: 'Species abundance'
 
+  - name: tool_anonymization
+    owner: ecology
+    tool_panel_section_label: 'Data exploration'
```
bgruening, line 76: they are ecology specific? Isn't it? Can we add them to one of the existing ecology categories, otherwise we need to adjust the filters for the ecology domain again ...

bgruening: Oh https://github.com/galaxyecology/tools-ecology/pull/47#issuecomment-898628686 haven't seen that.

If you wanna do that, can we name it "Biodiversity data exploration". Data exploration is really too generic.

PR #411, opened 2021-09-03
Update metabolomics.yaml

bgruening:
@lecorguille do you agree that we should auto-install all workflow4metabolomics?

### DIFF
--- a/metabolomics.yaml
+++ b/metabolomics.yaml
@@ -319,3 +319,10 @@ tools:
     owner: galaxyp
     tool_panel_section_label: Metabolomics
 
+
+  - name: isoplot
+    owner: workflow4metabolomics
+    tool_panel_section_label: Metabolomics
+
+
+


lecorguille:
I would be glad to have [tools-metabolomics](https://github.com/workflow4metabolomics/tools-metabolomics) in the trusty repositories :)

bgruening:
I can not add all tools to trusted, or can I? Many of those are not maintained or not reviewed, or did this changed? I would create a new workflow4metabolomics.yml similar to tools_iuc.yml where I put all the tools that are reviewed and in this repo. Is that plan ok?

lecorguille:
Almost all the tools are regularly updated. There aren't so much that are deprecated. 
To be sure, we can synchronise based on this [tool list](https://gitlab.com/ifb-elixirfr/usegalaxy-fr/tools/-/blob/master/files/workflow4metabolomics.yml).

bgruening:
Are those tools updated and reviewed? I'm just concerned about security stuff :(

lecorguille:
It's been a while that we integrate the `' '` around inputs and outputs but indeed we should check that they are really all ok.

lecorguille:
I will manage the checking or put someone on it.
Are we agree that the security issue are mainly around the protection in single quote of the in/outputs and the text param?

bgruening:
Yes, singe-quotes, sanitizing text inputs but also allowed input data. RData is bad and python pickles. That is the most important stuff.

lecorguille:
I was afraid about adding RData in the black list. The whole workflow msnbase -> xcms -> camera is as you may know based on RData.

PR #420, opened 2021-09-23
Add UMI tools

gallardoalba:


bgruening (approved):

bgruening:
have you not copied that from the other yaml file?

gallardoalba:
> have you not copied that from the other yaml file?

Not, those weren't included; I just realized because the UMI extract tool in Galaxy was outdated.

PR #424, opened 2021-10-13
Add tracy tools and Sanger Sequencing section

pvanheus:
Add the [tracy](https://github.com/gear-genomics/tracy) tools for working with Sanger sequencing chromatograms, also add a section for Sanger Sequencing (perhaps more Sanger sequencing related tools coming in the future).


bgruening:
Ok, lets go with this, thanks @pvanheus!

PR #438, opened 2021-11-15
Add hypo tool

jesgomez:


bgruening:
it should go to the tool_iuc.yml file, since its a IUC tool.

bgruening.yaml
```diff
@@ -232,3 +232,6 @@ tools:
     owner: bgruening
     tool_panel_section_label: Single-cell
 
+ - name: hypo
+   owner: bgruening
```
bgruening, line 236: and this should be `iuc`

jesgomez: yes, that makes sense. I just fixed it. Thanks

bgruening (approved):

bgruening:
Cool, thanks a lot. In 30 or so it will be installed!

PR #444, opened 2021-12-09
Add Dada2 and FROGS 3.2 on the list

yvanlebras:
dada2 already on the `Metagenomic Analysis` section of eu

### DIFF
--- a/ecology.yaml
+++ b/ecology.yaml
@@ -462,7 +462,14 @@ tools:
   - name: fastq_combiner
     owner: devteam
     tool_panel_section_label: 'FASTA/FASTQ'
-
+    
+  # FROGS
+  - name: frogs
+    owner: frogs
+    tool_panel_section_label: frogs
+    tool_panel_section_label: 'DNA Metabarcoding'
+    
+  # Obitools
   - name: obi_illumina_pairend
     owner: iuc
     tool_panel_section_label: 'DNA Metabarcoding'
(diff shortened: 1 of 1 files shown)


ecology.yaml
```diff
@@ -462,7 +462,14 @@ tools:
   - name: fastq_combiner
     owner: devteam
     tool_panel_section_label: 'FASTA/FASTQ'
-
+    
+  # FROGS
+  - name: frogs
```
bgruening, line 467: where is this tool developed? Does it fulfil our security standards?

yvanlebras: Hey ;) This tool is developed in France, notably by colleagues from INRAe. It is proposed through usegalaxy.fr and the dev repo is there https://github.com/geraldinepascal/FROGS-wrappers/ and it seems to me this was delevopped following good practices. Maybe we can ask further informations to @mariabernard or @lecorguille?

lecorguille: Recently, I enhanced the security by adding a lot of `'` but maybe I missed something  👁️ 👁️

mariabernard: I followed @lecorguille recommandations and I hope that everything respects the standards but if you have other suggestions I am all "ears".
The only point that we did not find a "perfect" solution is to automatically order FROGS tools in the tool pannel. @bgruening if you have any idea I am ready to follow it.
On the [readme of FROGS-wrappers](https://github.com/geraldinepascal/FROGS-wrappers#simplest-way) we suggest to modify tool_conf.xml.
Results of FROGS are also often HTML file, that need to not be sanitize ...

Let me know if you have comments.

bgruening: @mariabernard awesome. I wanted to look at the wrapper over the next days. One thing I found is the managing of the JAVA memory settings via a hidden parameter. I think this is not needed.
The ordering of tools in the panel can be controlled better with Galaxy 21.09. So there is hope :)

ecology.yaml
```diff
@@ -462,7 +462,14 @@ tools:
...
+    owner: frogs
```
bgruening, line 468: The owner is wrong and the section label is twice otherwise I think this is ready to go.

bgruening: If this is fixed I think we can merge it as soon as the new frogs version is in the toolshed.

bernt-matthias:
Just stumbled over this PR. Can we make this two PRs? One for dada2 which we probably can merge immediately and one for FROGS .. 

For FROGS I noticed that it includes some functionality of phyloseq and I would prefer when this would be syncronized with the recent efforts to add phyloseq to IUC: https://github.com/galaxyproject/tools-iuc/tree/master/tools/phyloseq

bernt-matthias:
Hmm. Why add dada2 again?

bgruening:
> Hmm. Why add dada2 again?

I think that is fine. It isicates that ecology needs those tools. Technically it die not make a difference.

bgruening:
Can this be closed?

bernt-matthias:
From my side yes.

yvanlebras:
Ok or me considering Dada2. For FROGS, it seems there is a new version now, I make a dedicated PR!

PR #447, opened 2022-01-21
Add climate.yaml for automatic update of tools

annefou:
- Update xarray tools to 0.20. 2 for Galaxy climate (if I understood properly, they won't be automatically updated).


bgruening:
Please add the climate YAML file to https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L37 and then they get automatically updated. I guess we forget to do that.

bgruening:
Do you need them today, or can we wait until Sunday?

annefou:
I am not in a hurry.

PR #450, opened 2022-02-11
Add read_it_and_keep - for decontaminating viral FASTQ reads

pvanheus:
This is a new tool for decontaminating (i.e. host removal etc) viral FASTQ reads.

I was not sure which section to put it in so I put it in the same section as FastqQC etc.

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -1371,6 +1371,10 @@ tools:
     owner: devteam
     tool_panel_section_label: FASTA/FASTQ
 
+  - name: read_it_and_keep
+    owner: iuc
+    tool_panel_section_label: FASTA/FASTQ
+
   - name: mitobim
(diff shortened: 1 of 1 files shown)


wm75:
https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/df023e59dd8e0c3c7c7dfc35149a5330a27800dd/tools_iuc.yaml#L3677-L3679
we've been faster, but thanks @pvanheus :)

bgruening:
its already included and should be installed ... its under virology, which is maybe wrong ...

https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/read_it_and_keep/read_it_and_keep/0.1.0+galaxy0

pvanheus:
Virology is right, thanks!

PR #453, opened 2022-03-29
Add MetaNovo tool.

reid-wagner:
Adds the MetaNovo tool.


gmauro:
It should be installed during the weekend

PR #459, opened 2022-04-29
Add coverm tools

gallardoalba:


gallardoalba:
This is the final required step for including Coverm in usegalaxy.eu @mira-miracoli

bgruening (approved):

tools_iuc.yaml
```diff
@@ -1599,6 +1599,14 @@ tools:
     owner: iuc
     tool_panel_section_label: Mapping
 
+  - name: coverm_contig
+    owner: iuc
+    tool_panel_section_label: Mapping
```
bebatut, line 1604: I am not it is the best section for CoverM. I would have put it in "Metagenomic Analysis"
In any case, thanks a lot for adding it. I will tell Magnus and Valerie that the tool is there

gallardoalba: You are right, sorry; I didn't check the CoverM README.

bgruening: Can you change the category here ... I will fix it on the server.

gallardoalba: Sure.

PR #466, opened 2022-05-27
Add orfipy

gallardoalba:


tools_iuc.yaml
```diff
@@ -948,6 +948,10 @@ tools:
     owner: iuc
     tool_panel_section_label: Extract Features
 
+  - name: orfipy
+    owner: iuc
+    tool_panel_section_label: Extract Features
```
bgruening, line 953: isn't that annotation?

gallardoalba: It generates FASTA and/or BED files, so I'm not sure if annotation fits.

bgruening (approved):

PR #502, opened 2022-09-21
Add new tools pathifier and gsc_scran_normalize from artbio based on …

pavanvidem:
…a user request


bgruening:
@pavanvidem have you checked briefly if those tools are "safe"?

pavanvidem:
They're from [artbio](https://github.com/ARTbio/tools-artbio). We already have some of their tools on EU server. Anyways, I checked them and they look good.

bgruening:
They will be installed at the weekend, is that enough?

pavanvidem:
Thanks!

PR #503, opened 2022-09-22
Create materials.yaml

elichad:


bgruening:
I think you need to add this new category here as well: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/.schema.yaml

elichad:
No auto-update for now as we're still figuring out the best update process for our tools. I've made a lockfile

bgruening:
Ok! They are now installed over the weekend. If you need to add. Ew versions you would need to add the version hash's to the lock file.

Thanks Eli!

PR #511, opened 2022-10-10
Add first batch of CPT phage annotation tools

jasonjgill:
Requested 11 tools needed for the structural annotation workflow, tools are all in the CPT-ToolshedSource repo. If tools can go into IUC that may be easier?


jasonjgill:
Check produces an error, one of the tool names in our repo differs from its name in the toolshed. Will try again,

PR #514, opened 2022-10-12
Install new tools: IUC, Goeckslab and Bgruening

gallardoalba:
Related: https://github.com/galaxyproject/bioblend/issues/462


gallardoalba:
One of the tools is not found in the toolshed, but I checked it and everything seems fine: `AssertionError: red not in toolshed`.

https://toolshed.g2.bx.psu.edu/view/iuc/red/18a46265455b

bgruening:
Here is my assumption. `red` is not very specific and you find many many hits. This line https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/scripts/pr-check.py#L21 returns paginated results and the red result it on page 3 or 4. So we need to enhance this script.

bgruening (approved):

bgruening:
Danke

PR #518, opened 2022-10-25
add dewseq + htseq-clip to galaxy

michauhl:
Add DEWSeq + htseq-clip to usegalaxy.eu. It would be great to have a new tool_panel_section_label "CLIP-seq" here, just like there is already one for "ChIP-seq" tools on usegalaxy.eu.

### DIFF
--- a/bgruening.yaml
+++ b/bgruening.yaml
@@ -298,8 +298,8 @@ tools:
 
   - name: dewseq
     owner: rnateam
-    tool_panel_section_label: CLIP-seq
+    tool_panel_section_label: Peak Calling
 
   - name: htseq_clip
     owner: rnateam
-    tool_panel_section_label: CLIP-seq
+    tool_panel_section_label: Peak Calling
--- a/bgruening.yaml
+++ b/bgruening.yaml
@@ -295,3 +295,11 @@ tools:
   - name: minipolish
     owner: bgruening
     tool_panel_section_label: Assembly
+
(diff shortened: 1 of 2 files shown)


bgruening:
Is a specific category needed? Can we not fold this into RNA?

bgruening:
Well if you think CLIP deserves its own category than ChIP ;) Then please add the new category to the schema file in this repo, so that the test does pass. Thanks.

michauhl:
I see, it's probably too much work (weird I thought there is a ChIP-seq category, but there is none). So let's just put them under "Peak Calling"

bgruening:
@michauhl ping

bgruening (approved):

bgruening:
Thanks.

michauhl:
Thank you Björn :) How about memory requirements of tools? Do you specify this in another file on here ?

bgruening:
We are currently migrating this, lets take the new one here: https://github.com/galaxyproject/tpv-shared-database/blob/main/tools.yml

@mira-miracoli can help you if needed I think.

michauhl:
I gathered the following requirements:
dewseq: 40GB RAM, 2 cores
htseq_clip: 16GB RAM, 4 cores
@mira-miracoli can you add these stats? I'm not sure how to specify these in https://github.com/galaxyproject/tpv-shared-database/blob/main/tools.yml

PR #535, opened 2022-12-09
update of gfastats version

Delphine-L:


bgruening:
This should be done automatically. Wired.

PR #543, opened 2023-01-13
Update ecology.yaml

yvanlebras:


bgruening:
> AssertionError: obisindicator not in toolshed.

ecology.yaml
```diff
@@ -471,6 +479,31 @@ tools:
...
+
+  - name: xarray_netcdf2netcdf
+    owner: ecology
+    tool_panel_section_label: 'GIS Data Handling'
+
+  - name: xarray_mapplot
+    owner: ecology
+    tool_panel_section_label: 'GIS Data Handling'
+
+  - name: bwa
```
bgruening, line 503: we can not have the same tool in multiple categories :(

Can we remove this? You would need to include the mapping section in your subdomain.

yvanlebras: yes, for sure! an error on my side I guess

PR #551, opened 2023-01-31
Add toulligqc to fasta/fastq section

PaulineF13:
Hello, could you add this tool ? 
toulligQC allows quality check for long reads data
Thks !

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -972,6 +972,10 @@ tools:
   - name: fasta_regex_finder
     owner: mbernt
     tool_panel_section_label:  FASTA/FASTQ
+    
+  - name: toulligqc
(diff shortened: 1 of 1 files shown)


bgruening:
@PaulineF13 is it this tool? https://github.com/GenomiqueENS/toulligQC

The tools you add here needs to be in the Galaxy toolshed, and it seems it is not. So we at first need to integrate this tool before we can install it.

PaulineF13:
> 

@bgruening Yes that's it and sorry, I didn't check the toolshed first, big mistake ! I will see with authors to add it, thanks

bgruening:
Let me know if they are not interested or if they need help, please. Then we try to find other ways :)

PR #554, opened 2023-02-09
add filter_tabular 3.3.0; preserve filter_tabular 2.0.0

eschen42:
@bgruening @hexylena 

This PR is to add `filter_tabular` to `tools_iuc.yaml`

Regarding the additions to `tools_iuc.yaml.lock`
`filter_tabular` 2.0.0 (revision 34d29339abab) is presently installed on UseGalaxy.eu
`filter_tabular` 3.3.0 (revision 4d5aae46f850) is to be installed onto UseGalaxy.eu

Thank you.

P.S. When I ran
```make fix```
it identified a number of issues that I did not include in this PR because it did not concern this tool.

### DIFF
--- a/tools_iuc.yaml.lock
+++ b/tools_iuc.yaml.lock
@@ -8366,7 +8366,6 @@ tools:
   owner: iuc
   revisions:
   - 4d5aae46f850
-  - 34d29339abab
   tool_panel_section_label: Text Manipulation
 - name: flanking_features
   owner: devteam
--- a/tools_iuc.yaml.lock
+++ b/tools_iuc.yaml.lock
@@ -8362,6 +8362,12 @@ tools:
   revisions:
   - 549d2cb4c6f2
   tool_panel_section_label: Text Manipulation
+- name: filter_tabular
+  owner: iuc
+  revisions:
+  - 4d5aae46f850
+  - 34d29339abab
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tools_iuc.yaml.lock [hunk 1]
```diff
@@ -8362,6 +8362,12 @@ tools:
   revisions:
   - 549d2cb4c6f2
   tool_panel_section_label: Text Manipulation
+- name: filter_tabular
+  owner: iuc
+  revisions:
+  - 4d5aae46f850
```
bgruening, line 8368: do you need both version? can we just take the latest?

bgruening: ping @eschen42

eschen42:
@bgruening I removed the filter_tabular 2.0.0 version (which is presently installed on usegalaxy.eu) and left only the 3.3.0 version (which I need), assuming that this won't uninstall the old version and break others' workflows.  I also added v0.1.18 of galaxyp mqppep_anova and mqppep_preproc *without* removing older versions (because scripts/update-tool.py doesn't do that).

Is this what you expect?

Thank you.

bgruening (approved):
Yes, this looks correct.

eschen42:
> Yes, this looks correct.

Thank you @bgruening

PR #558, opened 2023-02-17
Add metabolomics to trusted

gallardoalba:


bgruening:
I'm not sure this is a good idea, many of those tools are older and maybe insecure, last time I checked. Can we bump them case by case or follow whatever @abretaud is doing?

gallardoalba:
> 

Ok, I'll close then :)

PR #574, opened 2023-04-25
Add aligned_to_consensus tool

ColineRoyaux:


ecology.yaml
```diff
@@ -775,3 +775,7 @@ tools:
   - name: ab1_fastq_converter
     owner: ecology
     tool_panel_section_label: 'Convert Formats'
+    
+  - name: aligned_to_consensus
```
bgruening, line 779: Are you sure the tool is named like that in the TS? It is not available?

ColineRoyaux: It was right actually

bgruening:
It will be installed over the weekend, let me know if you need it sooner.

ColineRoyaux:
Thank you so much Bjorn ! It's not urgent don't worry :)

PR #577, opened 2023-05-03
Update tools_iuc.yaml

paulzierep:
add deepmicro

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -1475,6 +1475,10 @@ tools:
 
 # MACHINE LEARNING
 
+  - name: deepmicro
+    owner: iuc
+    tool_panel_section_label: Machine Learning
+
   - name: mcl
     owner: iuc
     tool_panel_section_label: Machine Learning


tools_iuc.yaml
```diff
@@ -1475,6 +1475,10 @@ tools:
 
 # MACHINE LEARNING
 
+  - name: deepmicro
+    owner: iuc
+    tool_panel_section_label: Machine Learning
```
bgruening, line 1480: It it really Machine Learning? The fact that it uses a certain method is a detail that is not interesting for the user, isn't it?

paulzierep: despite its name it really is just an autoencoder technique, it could therefore be used for any kind of features - not limited  to microbiome data at all ... ? therefore, I thought ML would match best. I am thinking about a bachelor/master project where we can make it a bit more generic, i.e. allowing the export of the model and transform on new data, then I would probably rename it just to `autoencoder` ... but I thought it would be fair to name it deepmicro atm to give credit to the original publication ...

paulzierep:
Can I merge it ? Or should I move to `# METAGENOMIC ANALYSIS` ?

bgruening (approved):

bgruening:
Sorry, forgot about it.

PR #578, opened 2023-05-03
Remove humann with old versioning (Humann2)

tuncK:
Else, it pops up in the search box as if it is a separate tool.


bgruening:
This will not uninstall tools, we do not uninstall tools at all and don't have a procedure for it.

I can hide tools if you like.

tuncK:
Hmmm, OK. Then yes, can you please hide humann2*?

bgruening:
Can you see if I have hidden all in 10min?

tuncK:
Yes, hiding worked as of today 12:27. Thanks

bgruening:
Can we close this?

PR #579, opened 2023-05-03
Add PepQuery2 tools

reid-wagner:
Adds `pepquery2`, `pepquery2_index`, and `pepquery2_show_sets` under the Proteomics label.

I gleaned that adding a separate entry for each tool was preferred to an entry for the `suite_pepquery2` suite, but please let me know if that's not correct.


bgruening (approved):

bgruening:
Thanks it will be installed over the weekend. If you need it earlier let me know.

reid-wagner:
Thank you!

PR #583, opened 2023-05-16
Add CNVkit

khaled196:
Add CNVkit


bgruening:
Its failing with "cnvkit_coverage" is not in the TS ... did we make a mistake :)

khaled196:
Hi @bgruening 

I am sorry for that. The tool cnvkit_coverage disappeared from the .git and was ignored from the git add after that. I have updated the tools again.

PR #590, opened 2023-06-07
Add manta

hexylena:
For @scottcain, and, maybe also interesting for @wm75 and my coworkers.

- tool looks ok, artbio is good
- they were careful about quoting https://github.com/ARTbio/tools-artbio/blob/master/tools/manta/manta.xml

should be ok :+1: 
pinging @bgruening (is one of your admin team now handling tool requests, should I ping someone else?)

### DIFF
--- a/genome-annotation.yaml
+++ b/genome-annotation.yaml
@@ -123,3 +123,7 @@ tools:
   - name: tbl2gff3
     owner: iuc
     tool_panel_section_label: Annotation
+    
+  - name: manta
+    owner: artbio
+    tool_panel_section_label: Variant Calling


bgruening:
I found one script that was unquoted and added a profile version here: https://github.com/ARTbio/tools-artbio/pull/618

bgruening:
@hexylena maybe you can ping the @usegalaxy-eu/admin group?

wm75:
arrgh, https://github.com/galaxyproject/tools-iuc/pull/3140 fell victim to covid, sorry.
I don't remember at all anymore what the status was, nor what I thought of the existing wrapper.

hexylena:
@bgruening yes, can do in the future :) @wm75 ahhh jeez.

hexylena:
> I found one script that was unquoted and added a profile version here: 

looks like just the python file in the tool directory in your PR, so, at least not a security risk, albeit sub-optimal.

bgruening:
I will merge. @wm75 maybe we can migrate this one with this user to IUC if you ever find time to work on it.

hexylena:
Thanks @bgruening , @wm75 !

PR #592, opened 2023-06-15
Add RECETOX tools

xtrojak:
First wave of adding RECETOX tools to usegalaxy.eu.
This PR brings tools `recetox-aplcms`, `MSMetaEnhancer`, `RIAssginer`, `RAMClustR`, `Biotransformer`, and `RMassBank` to Metabolimics section.

### DIFF
--- a/eirene.yaml
+++ b/eirene.yaml
@@ -1,0 +1,52 @@
+---
+install_repository_dependencies: true
+install_resolver_dependencies: true
+install_tool_dependencies: false
+tools:
+  - name: recetox_aplcms_generate_feature_table
+    owner: recetox
+    tool_panel_section_label: Metabolomics
+
+  - name: recetox_aplcms_recover_weaker_signals
+    owner: recetox
+    tool_panel_section_label: Metabolomics
+
+  - name: recetox_aplcms_correct_time
+    owner: recetox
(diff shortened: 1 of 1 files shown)


eirene.yaml
```diff
@@ -0,0 +1,52 @@
+---
+install_repository_dependencies: true
+install_resolver_dependencies: true
+install_tool_dependencies: false
+tools:
+  - name: recetox_aplcms_generate_feature_table
+    owner: recetox
+    tool_panel_section_label: EIRENE
```
bgruening, line 8: Do users really know what EIRENE is and will find tools in this category? Maybe a more scientific label would be more useful?

hechth: I think the whole Exposome debate is too heated to classify tools as "Exposomics" only, so I think having an EIRENE section is cleaner at the moment. We will add Exposomics to the Ontology topics and maybe also to the toolshed so that we can classify the tools there? this should make sure they are findable.

bgruening:
@xtrojak we install tools over the weekend, do you need them before?

xtrojak:
> @xtrojak we install tools over the weekend, do you need them before?

The weekend is completely fine. Thank you!

PR #603, opened 2023-07-07
Added eirene.yaml to list of trusted tools

hechth:
Only tools mature enough for auto-update will be included in the list of tools published in the yaml file.


bgruening (approved):
Many thanks. 

grafik

PR #608, opened 2023-07-19
Update materials tools for MSS 2.2.1 and PMS 0.3.0

patrick-austin:
Adds utility tool for MuSpinSim, and version updates for most `muon-spectroscopy-computational-project` tools.

### DIFF
--- a/materials.yaml
+++ b/materials.yaml
@@ -16,6 +16,10 @@ tools:
     owner: muon-spectroscopy-computational-project
     tool_panel_section_label: Muon Spectroscopy
 
+  - name: muspinsim_combine
+    owner: muon-spectroscopy-computational-project
+    tool_panel_section_label: Muon Spectroscopy
+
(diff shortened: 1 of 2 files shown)


bgruening:
@patrick-austin did I simply forget this one? Or was it not ready? Sorry if we just missed it. Please ping us in those cases.

patrick-austin:
There was an issue with `pm-uep-opt` that needing patching with `0.3.0+galaxy1` but now it should be ready to merge. I think I opened this the week after the Australia conference so probably a lot of other things going on at the time.

bgruening:
Do we need to update the loc files? Or is it up to date?

patrick-austin:
`materials.yaml.lock` has all the new revision ids for the tools that have changed and a new entry for `muspinsim_combine` - I think that's everything that needs to change? So it should be up to date as is.

bgruening (approved):

PR #618, opened 2023-09-04
Astronomical Archives (IVOA) Data Source tools

volodymyrss:
Astronomical Archives (IVOA) Data Source tools

https://toolshed.g2.bx.psu.edu/view/astroteam/astronomical_archives/0ddfc343f9f9

developed by @francoismg

### DIFF
--- a/astronomy.yaml
+++ b/astronomy.yaml
@@ -15,3 +15,7 @@ tools:
   - name: astropy_fits2bitmap
     owner: astroteam
     tool_panel_section_label: 'Astronomy'
+
+  - name: astronomical_archives
+    owner: astroteam
+    tool_panel_section_label: 'Astronomy'


francoismg:
I can find the tool when searching the toolshed from my galaxy instance, but some things seems weird, will upload a screenshot

volodymyrss:
> I can find the tool when searching the toolshed from my galaxy instance, but some things seems weird, will upload a screenshot

I imagine toolshed instability might be the cause. But then this PR here managed to find and test the tool, right?

francoismg:
Screenshot 2023-09-04 at 16 45 40
Screenshot 2023-09-04 at 16 46 52


> > I can find the tool when searching the toolshed from my galaxy instance, but some things seems weird, will upload a screenshot
> 
> I imagine toolshed instability might be the cause. But then this PR here managed to find and test the tool, right?

yes it seems like it was able to do so since it uploaded it to the tool shed and maybe the remaining problem are just caused by some instability like you said

volodymyrss:
> But I'm not sure about the fact that there's an "astroteam" version and a "volodymyr.savchenko" version, do you have that too?

Yes, I changed it in the .shed.yml, it should be astroteam, but first version went with my account which is not right.

Does it work to download the astroteam one?

francoismg:
> > But I'm not sure about the fact that there's an "astroteam" version and a "volodymyr.savchenko" version, do you have that too?
> 
> Yes, I changed it in the .shed.yml, it should be astroteam, but first version went with my account which is not right.
> 
> Does it work to download the astroteam one?

yes the astroteam one seems to work

bgruening (approved):

bgruening:
Its installed now.

PR #622, opened 2023-09-17
Update ecology.yaml Add EAM Assembly line tools

yvanlebras:
Need to add a new usegalaxy.eu section "Data and Metadata Management"

### DIFF
--- a/.schema.yaml
+++ b/.schema.yaml
@@ -24,7 +24,7 @@ mapping:
                 "tool_panel_section_label":
                     type: str
                     enum: [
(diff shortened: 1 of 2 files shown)


bgruening (approved):
I see, it need to add this section. Are you sure it does not fit into any other section?

yvanlebras:
We can use existing section BUT this will not fit 100% and it seems to me important to have such section where we can put rdm Galaxy tools . Related to Galaxy Elixir IS. I have several tools to put in this section;) BUT please, don't hesitate to say if not so relevant!

yvanlebras:
THANK YOU !!!

PR #628, opened 2023-10-05
Update eirene.yaml.lock file with versions required for GCMS training

hechth:
@bgruening should these not somehow be automatically updated? For some reason, the training apparently contains versions which are not on EU, eventhough we didn't update the training since we brought all the tools to EU if I am not mistaken.

### DIFF
--- a/eirene.yaml
+++ b/eirene.yaml
@@ -102,3 +102,7 @@ tools:
   - name: matchms_split
     owner: recetox
     tool_panel_section_label: Metabolomics
+
+  - name: waveica
+    owner: recetox
+    tool_panel_section_label: Metabolomics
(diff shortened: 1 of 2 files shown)


bgruening:
Its not currently in the trusted tool list ;)
Maybe we should change that.

hechth:
> Its not currently in the trusted tool list ;) Maybe we should change that.

I thought we made that change some time ago? Or do you mean only the waveica tool?

bgruening:
True: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L54

Ok, when have those tools be updated? We only update them once a week.

bgruening (approved):

PR #630, opened 2023-10-10
add dimet tools in new Metabolomics section

bdartigues:


bgruening:
@bdartigues great work! Really cool! Thanks a lot. It will be online next Monday.

Do you want to write a small news item so we can point our users to those new tools?

bdartigues:
@bgruening,
Once again, I'd like to thank the whole galaxy team for your advice and your sharp vision of our tools and the improvements you make to them.  It's great for our tool to have the visibility that galaxy brings. Okay, I'll check with Johanna and Macha to prepare a short summary of the tool. 
Thanks a million again!

PR #647, opened 2023-11-14
updates imaging tools to latest revisions

kostrykin:
Reflects changes made by following PRs:
- https://github.com/BMCV/galaxy-image-analysis/pull/92
- https://github.com/BMCV/galaxy-image-analysis/pull/91

The tool `2d_split_binaryimage_by_watershed` is removed (functionality is now merged into the `binary2label` tool).

### DIFF
--- a/imaging.yaml.lock
+++ b/imaging.yaml.lock
@@ -10,6 +10,7 @@ tools:
   owner: imgteam
   revisions:
   - 589af0005df5
+  - b74693340624
   - bf590a9733ed
   tool_panel_section_label: Imaging
 - name: color_deconvolution
--- a/imaging.yaml.lock
+++ b/imaging.yaml.lock
@@ -77,6 +77,7 @@ tools:
 - name: rfove
   owner: imgteam
   revisions:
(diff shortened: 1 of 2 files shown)


imaging.yaml.lock
```diff
@@ -158,12 +159,6 @@ tools:
   - d0960e1b25a8
   - f8f1100d0701
   tool_panel_section_label: Imaging
-- name: 2d_split_binaryimage_by_watershed
```
bgruening, line 161: this will not remove the tool, please leave it in. We need to hide this tool from the tool panel.

We need to keep it installed forever, to not break workflows.

kostrykin: Like that? d6c007a

bgruening: No please, just don't touch the tools that needs to be deprecated/removed. We can not do this in this repo yet.

kostrykin: Got it, 057f900 should be what we need then.

How to remove that tool from the tool panel?

PR #656, opened 2023-11-22
adds newest revision of `bioformats2raw` tool

kostrykin:
xref https://github.com/BMCV/galaxy-image-analysis/pull/99


bgruening:
https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L40C1-L40C45

I don't think you need to do that. On Saturday every week we should get the latest revision.

kostrykin:
Ah, didn't know that! 👍

PR #664, opened 2023-12-19
Add bbtools_tadpole, SeqSero2, and RGI

WhoisDonlee:
The RGI wrapper and data manager is provided by card, which is also de developer of RGI. I wasn't sure if this was allowed so let me know if it belongs somewhere else.

I didn't run into any problems running it on our own instance.

https://github.com/arpcard/rgi_wrapper/tree/master

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -4812,9 +4812,6 @@ tools:
   - name: data_manager_mapseq
     owner: iuc
 
-  - name: rgi_database_builder
-    owner: card
-
 # DEEPTOOLS
 
   - name: deeptools_alignmentsieve
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -4812,6 +4812,9 @@ tools:
   - name: data_manager_mapseq
     owner: iuc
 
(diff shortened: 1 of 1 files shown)


bgruening:
@WhoisDonlee rgi_database_builder does not seem to be in the TS.

The others are not a problem to install, but the rgi tools need a review or moved to IUC.

WhoisDonlee:
The rgi_database_builder is installed when the rgi_wrapper is installed via the admin panel. I'm not that familiar with data managers, but it looks like it is included in the rgi_wrapper itself and not as a separate tool.

bgruening:
Yes, which is also not good. I think we need to work in the rgi a bit. Do you know the author?

WhoisDonlee:
Ah okay, for now I will remove it and will make it a separate PR once it's in the IUC.
Thanks!

bgruening (approved):

bgruening:
Supi! This will be installed automatically on Sunday. Let me know if you need it before Xmas :)

WhoisDonlee:
Awesome! Sunday will be fine!

PR #686, opened 2024-03-09
Add `image_math` tool

kostrykin:
Add the `image_math` tool from https://github.com/BMCV/galaxy-image-analysis/pull/107

### DIFF
--- a/imaging.yaml
+++ b/imaging.yaml
@@ -308,6 +308,10 @@ tools:
     owner: imgteam
     tool_panel_section_label: "Imaging"
 
+  - name: image_math
+    owner: imgteam
+    tool_panel_section_label: "Imaging"
+
   - name: cp_cellprofiler
     owner: bgruening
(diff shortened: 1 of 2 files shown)


bgruening (approved):
For the imgteam tools we do the update of the lock files automatically. https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L40C1-L40C45

So no need to modify this file for trusted tools.

kostrykin:
> For the imgteam tools we do the update of the lock files automatically. https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/Makefile#L40C1-L40C45
> 
> So no need to modify this file for trusted tools.

Ah okay, this means I still need to add the entries for new tools to the .lock files, but not their revisions, right?

bgruening:
No, don't touch the lock file, its not needed for your tools

PR #702, opened 2024-04-02
Revision for Helixer

rlibouba:
Hello, 
A new version of Helixer is available. The PR for this version has been merged. I'm opening this PR to add it to the revision section so that it can be used on usegalaxy.eu.

Thank you!
Have a good day!


bgruening:
Do we need to change also TPV?

PR #704, opened 2024-04-05
Add pairtools for processing sequencing data from a Hi-C experiment

SaimMomin12:
Added 5 tools from the pairtools suite

- pairtools_parse
- pairtools_dedup
- pairtools_split
- pairtools_sort
- pairtools_stats


bgruening:
Thanks once added they will be installed and updated every weekend.

PR #705, opened 2024-04-13
added chewbbaca to tools_iuc.yaml

nilchia:


tools_iuc.yaml
```diff
@@ -3194,6 +3194,10 @@ tools:
 
 # PHYLOGENETICS
 
+  - name: chewbbaca
```
bgruening, line 3197: That is wrong. You need to add all tools in the toolshed. You created a lot of different tools https://toolshed.g2.bx.psu.edu/

nilchia: Oh, thanks. I thought it only needs the tool name mentioned in the .shed.yml file.
Now I added all 8 tools.

PR #718, opened 2024-05-15
add virheat to usegalaxy.eu

PlushZ:


wm75:
@PlushZ @bgruening maybe here Virology is the better section.
I know the old Variant Frequency Plot is in Variant Calling but that wasn't the best decision and I never remember.

bgruening:
Ok, can you move both tools into the new section?

PR #719, opened 2024-05-15
move virheat and snpfreqplot to virolgy 

PlushZ:


bgruening:
@wm75 you meant Virology should be used isn't it? https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/718

tools_iuc.yaml
```diff
@@ -4966,11 +4962,20 @@ tools:
...
+  
+  - name: snpfreqplot
+    owner: iuc
+    tool_panel_section_label: Virology
 
   - name: varvamp
     owner: iuc
     tool_panel_section_label: Virology
 
+  - name: virheat
```
wm75, line 4974: I guess, you need to *move* virheat since your first PR is merged already.

wm75:
@bgruening is moving snpfreqplot (an already installed tool) possible this way?

If yes that's nice, if not I don't care much. Once virheat is tested and fully functional, we will probably hide the old tool anyway.

bgruening:
Yes. But I need to do something manually which is ok. Better to have it consistent.

PR #731, opened 2024-06-05
Add refseq_masher to assembly analysis

clsiguret:
It will be used for the 'Bacterial genome assembly' workflow as mentioned is this [PR](https://github.com/galaxyproject/iwc/pull/438)

Thanks !


bgruening:
Thanks @clsiguret !

PR #734, opened 2024-06-06
AddF FROGS Metabarcoding suite

yvanlebras:
Related to https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/444

If FROGS is ok now to be installed on eu.

### DIFF
--- a/ecology.yaml
+++ b/ecology.yaml
@@ -886,4 +886,7 @@ tools:
   - name: sdmpredictors_list_layers
     owner: ecology
     tool_panel_section_label: 'Get Data'    
-    
+
+  - name: frogs
+    owner: frogs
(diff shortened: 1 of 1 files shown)


bgruening:
@yvanlebras close?

yvanlebras:
Thank you Björn for coming back to me on that PR. Seems to me FROGS toolsuite is not on eu. An official up to date repo seems to be on the toolshed https://toolshed.g2.bx.psu.edu/repositories/11a27cab761b5c8c . Do you think ok to install it on eu ? I think here I was only updating the ecology.yaml file without asking to install frogs toolsuite before. I can do it if you are ok with that.

bgruening:
Well, will you use it? :)

And its a single repo, with multiple tools, so not really best practice atm, isn't it?

PR #752, opened 2024-07-11
Add infer_rnaformer

ivelet:


bgruening:
There is a specific file for rnateam, can you please use that instead?

PR #753, opened 2024-07-12
Add mg_toolkit_original_metadata

RZ9082:


tools_iuc.yaml
```diff
@@ -2405,6 +2405,10 @@ tools:
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
 
+  - name: mg_toolkit_original_metadata
```
bgruening, line 2408:
```suggestion
  - name: mg_toolkit
```

bgruening:
@RZ9082 does this look corect?

RZ9082:
> @RZ9082 does this look corect?

This is the folder yes. But i thought it has to be the tool id.

bgruening:
Strictly is has to be the ToolShed repository name: https://github.com/galaxyproject/tools-iuc/blob/main/tools/mg_toolkit/.shed.yml#L1

Unfortunatately, you have this as the repo name here:
https://github.com/galaxyproject/tools-iuc/blob/main/tools/mg_toolkit/.shed.yml#L1

Which is not the same as the tool-ID

tools_iuc.yaml
```diff
@@ -2405,6 +2405,10 @@ tools:
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
 
+  - name: mg_toolkit
```
bgruening, line 2408:
```suggestion
  - name: mg_toolkit_original_metadata
```

bgruening (approved):

PR #754, opened 2024-07-16
Add XCMS plotting tools to eirene.yaml

hechth:
This includes the tool XCMS based plotting tools added to tools-metabolomics added recently.


bgruening (approved):

hechth:
@bgruening it seems like the tools didn't get installed - can it be related to the fact that they have a different owner?

PR #757, opened 2024-07-18
Update goeckslab.yaml squidpy

nilchia:


goeckslab.yaml
```diff
@@ -17,10 +17,14 @@ tools:
     owner: goeckslab
     tool_panel_section_label: Spatial Omics
 
-  - name: squidpy
+  - name: squidpy_scatter
```
bgruening, line 20: yes, this is unfortunately true. I can hide the older version with the old ID ... if that is ok for you. You would need to check manually that no workflow of yours contain the old version.

pavanvidem: We are not using it so far in any workflow. Maybe Goecks' lab members are using it. But hiding is fine.

PR #763, opened 2024-07-30
add mudirac, remove deprecated tools

subindev-d:
removed the deprecated tool
added new tool - mudirac

### DIFF
--- a/bgruening.yaml.lock
+++ b/bgruening.yaml.lock
@@ -90,12 +90,14 @@ tools:
   revisions:
   - 6d8444408ff1
   - a6318f87f2cd
+  - abba05ead5d7
   - e009d8260be2
   tool_panel_section_label: Annotation
 - name: cpat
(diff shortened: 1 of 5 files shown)


bgruening:
@subindev-d we normally don't remove tools for reproducibility reasons. We can hide them or add a deprecation badge to it if you like.

On the other hand its your community and you decide if reproducibility is important and if old workflows should still be able to run.

What do you think?

bgruening:
Ok, please don't remove them from the list and provide me a list with tool IDs that I need to hide.

subindev-d:
> Ok, please don't remove them from the list and provide me a list with tool IDs that I need to hide.
@bgruening , these three are the tool to be hided
- pm_dftb_opt
- pm_muairss_write
- pm_yaml_config

bgruening:
Ok, should be hidden tomorrow.

PR #774, opened 2024-08-26
Add multiple tools [Taxonkit name2taxid, gtdb_to_taxdump and biobox_add_taxid]

SantaMcCloud:
Tools for CAMI Amber such that it can be fully used now.

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -1820,6 +1820,10 @@ tools:
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
 
+  - name: name2taxid
+    owner: iuc
+    tool_panel_section_label: Metagenomic Analysis
+
   - name: krakentools_kreport2krona
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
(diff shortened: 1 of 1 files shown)


tools_iuc.yaml
```diff
@@ -1820,6 +1820,18 @@ tools:
     owner: iuc
     tool_panel_section_label: Metagenomic Analysis
 
+  - name: name2taxid
+    owner: iuc
+    tool_panel_section_label: Metagenomic Analysis
+
+  - name: ncbi_gtdb_map
```
bgruening, line 1827:
```suggestion
  - name: gtdb_to_taxdump
```

bgruening: @SantaMcCloud this is what you mean, correct?

You are correct in the sense that the `<tool id=""` should be `gtdb_to_taxdump` as well, but it isn't. Should we fix that ...

SantaMcCloud: Yes this is what i mean. I currently not home so i can fix it later aswell if it not urgent. Should i change the wrapper if and the `.shed` name or obly the wrapper if to `gtdb_to_taxdump`? Also if you want to fix it you can also do i when you have room for it!

PR #786, opened 2024-10-01
Add RGI

hexylena:


tools_iuc.yaml
```diff
@@ -5211,6 +5211,9 @@ tools:
     owner: iuc
     tool_panel_section_label: Virology
 
+  - name: rgi # Includes a DM.
```
bgruening, line 5214: Thanks Helena. This is not an IUC tool and should go into some other file I guess. We might also need to check every version - so not auto-update this one?

This also looks a bit non-standard and I'm wondering if there is more to setup up.

```
#if $db_opts.db_opts_selector == "default":
    export DB_PATH='${__tool_data_path__}/rgi_databases/_default' &&
    export DATA_PATH="\$_GALAXY_JOB_TMP_DIR" &&
    mkdir -p "\$DB_PATH" &&
#end if
```

hexylena: Oh, you're right @ non-iuc, and agreed at not auto-updating. Do you have a preference for which file?

> I'm wondering if there is more to setup up.

I've run this in my internal galaxy, and seems to run fine once the included DM's been run.

bgruening: Yes, the DM (which also should be in a separate repo) is working but there is a 3rd option which is "default" and I'm not sure how this should work.

hexylena: Yes, I guess the default option would not work.

hexylena: (Did you intend to paste the same comment again?)

bgruening: No, just bad internet at the conference ...

bgruening: How should we proceed? Do you know the author of the wrapper?

hexylena: I do not know the author of the wrapper

hexylena:
I will close it for now. it isn't enough of a priority, and I do not have time to move their tools into iuc/negotiate with them/rewrite their tool to not use a strange local option.

bgruening:
If you need the tool we can work on that and try to get it into IUC for you.

PR #790, opened 2024-10-14
Add aquainfra processes

MarkusKonk:


ecology.yaml
```diff
@@ -895,6 +895,10 @@ tools:
     owner: ecology
     tool_panel_section_label: 'Get Data'
 
+  - name: aquainfra
```
bgruening, line 898: Oo, was that really the tool ID? This is way to generic :(

MarkusKonk: Yeah, I also thought that this morning, but I thought it's too late to change that...

MarkusKonk: @bgruening  Shall I just update the tool with a more specific id?

bgruening: until now its not installled please change that ...

bgruening:
Wann brauchst du es?

MarkusKonk:
Ich brauche das nicht zwingend diese Woche. Wenn das Samstag online geht, reicht das wohl.

PR #800, opened 2024-11-08
Added XAS tools

subindev-d:
Added new XAS tools


bgruening:
Please add the new sections here: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/.schema.yaml

bgruening:
Thanks @subindev-d

patrick-austin:
Hi @bgruening, I was looking on https://materials.usegalaxy.eu/ and the XAS tools don't appear (either as a section or via search). I don't know if it's the problem, but I noticed that that our `tool_panel_section_id` is `xas` but the `tool_panel_section_label` is `XAS (X-ray Absorption Spectroscopy)` whereas most tools have the id as a lowercase, underscored version of the label? I'm not sure where else to look for debugging.

bgruening:
@patrick-austin yes, sorry we are not there yet. We didn't have time so far to simplify the process. You would need this change for now: https://github.com/usegalaxy-eu/infrastructure-playbook/commit/738e18dd7648a33bc617e2cc8c56fb2deac9e36b

bgruening:
Should be fixed tm.

PR #802, opened 2024-11-28
Update tools_iuc.yaml by adding groot

hugolefeuvre:
Add groot tool into galaxy eu


bgruening (approved):
Merci!

bgruening:
This will be automatically installed now over the next weekend and will update then every weekend when new version are released.

PR #819, opened 2025-01-21
Add tools from the new EMBOSS v6 suite

shiltemann:
@bgruening I added both needleall and needle, but needle also has a v5 wrapper installed on EU currently, so not sure how to handle that to minimize confusion, do we deprecate the old version, make a different EMBOSS6 toolbox section?

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -396,7 +396,7 @@ tools:
   - name: spotyping
     owner: iuc
     tool_panel_section_label: Annotation
-  
+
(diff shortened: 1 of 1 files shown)


bgruening:
I would hide the old version. You could still use them in a workflow. Would that be ok?

bgruening:
I mean I can hide individual tools.

bgruening (approved):

shiltemann:
sounds good to me, thanks!

PR #820, opened 2025-01-22
add spacexr

nilchia:
I added spacexr under single-cell section.
Can I create a new "spatial" section?

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -4438,6 +4438,13 @@ tools:
     owner: iuc
     tool_panel_section_label: Single-cell
 
+  - name: spacexr_rctd
+    owner: iuc
+    tool_panel_section_label: Single-cell
+  
+  - name: spacexr_cside
+    owner: iuc
+    tool_panel_section_label: Single-cell
 # STATISTICS
 
   - name: best_regression_subsets


nilchia:
why is there only "suit_spacexr" on the tool shed?
 https://toolshed.g2.bx.psu.edu/view/iuc/suite_spacexr

pavanvidem:
@nilchia please add `auto_tool_repositories`to the .shed.yml file here https://github.com/galaxyproject/tools-iuc/blob/main/tools/spacexr/.shed.yml

bgruening:
We could add "spatial" but the overlap will be large with single-cell, isn't it? Should we maybe rename Single cell to "Spatial and Single-Cell ..."

pavanvidem:
there is already a Spatial Omics TS category

bgruening:
We are not talking about TS categories. Tools in the TS can be in multiple categories, but this is not possible in Galaxy

pavanvidem:
Let's keep both the categories for now and spacexr under the Spatial Omics category.

nilchia:
Sorry, I thought it was merged.
Should I move it to spatial_omics.yaml file?

bgruening:
No, they are IUC tools and they should stay in IUC

bgruening (approved):

nilchia:
Thanks!

PR #831, opened 2025-02-21
Add new tools-iuc clair3 DM

wm75:


bgruening:
Thanks. I already installed it and run it ...

PR #833, opened 2025-02-28
Tools sync workflow

pauldg:
Want to test locally with a tool like [act](https://github.com/nektos/act) first but it is not working yet (Docker engine problems). 

Closes #39.


bgruening:
@pauldg can we help here? This seems to be a very good idea!

PR #838, opened 2025-03-20
Add script to remove uninstallable revisions from lock files

bernt-matthias:


scripts/fix-uninstallable.py
```diff
@@ -0,0 +1,127 @@
+# check all revisions in the lockfile if they are installable
+# remove if not and the next installable version is installed
```
bgruening, line 2: But this script is not installing, isn't it? will it add the next proper version to the yaml file?

scripts/fix-uninstallable.py
```diff
@@ -0,0 +1,127 @@
...
+            nxt = None
+            for i in range(start, len(all_revisions)):
+                if all_revisions[i] in ordered_installable_revisions:
+                    nxt = all_revisions[i]
+            if nxt:
+                if nxt in tool["revisions"]:
+                    print(f"remove {cur} in favor of {nxt} {name} {owner}")
+                    remove.append(cur)
+                else:
+                    # not adding, will be done in next update anyway
```
bernt-matthias, line 102: I decided not to add the next installable one, because this is already covered by `update-tool.py`. On my CI I added it to the Makefile: https://github.com/Helmholtz-UFZ/ufz-galaxy-tools/blob/3dd6906f3f7a9641d96e3f7f7e5e4ee57becea53/Makefile#L27

martenson: Doesn't `update tool` add the latest, and not the next one?

If I have this repo history:

- aaaa (v1)
- bbbb (v1)
- cccc (v2)

and my lockfile has `aaaa`, I want to not just remove not-installable `aaaa`, I also want to add `bbbb`, because update-tools adds `cccc`.

Am I wrong here?

bernt-matthias: Yeah, I can add it, seems trivial. Good thing as that by now nothing is removed. 

Since I'm running this already in my CI, here are some stats: 

Out of the 4302 revisions listed in my lock file 622 have been removed in the [first run](https://github.com/Helmholtz-UFZ/ufz-galaxy-tools/commit/10295ab425417e08b4590cff4e40d657db3cb39d), i.e. more than 10%. In 5 cases we ended up in this else branch. 

Just want to highlight the impact this has .. I still think that your suggestion is good.

Also stumbled over some removals in the second round which I did not expect which tought me a bit about a "shortcomming" of the suite concept:

```
remove bbb4a531d80b in favor of dcbe3affb349 ampvis2_venn iuc
remove e85ff12bc025 in favor of dcbe3affb349 ampvis2_venn iuc
remove f28f2faad83b in favor of dcbe3affb349 ampvis2_venn iuc
```

In the three cases in each revision test-data was updated for other tools of the suite, which also created a new revision for this tool as well (which I wasn't aware of before). All this is perfectly fine - just want to share the insight. 

What is still missing is a script uninstall the revisions now absent from the lock files from the Galaxy servers.

bernt-matthias: I see, I can check.

bgruening:
Hi guys, I lost a but track here. I'm happy to merge it and keep the script here, but please feel also free to create a run-PR of this script so we see the impact and to clean the repo.

Thanks

scripts/fix-uninstallable.py
```diff
@@ -0,0 +1,163 @@
...
+            repo_path,
+        ]
+        proc = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    else:
+        cmd = ["hg", "pull", "-u"]
+        proc = subprocess.run(cmd, cwd = repo_path, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    assert proc.returncode == 0, f"failed {' '.join(cmd)}"
+
+def get_all_revisions(toolshed_url: str, name: str, owner: str) -> List[str]:
+    repo_path = f"/tmp/repos/toolshed-{owner}-{name}"
```
martenson, line 59: with other toolsheds in place this could collide

bernt-matthias:
Two more comments: 

- It seems that there can be uninstallable revisions that are not due to having the same version as a later revision (e.g. flye from bgruening de24438c9988 where the next installable revision would be 8d4f03b5fe9d, graphclust_align_cluster has no installable revisions). No idea what is the reason for this. But I changed the code such that there is only a warning in such cases. 
- I found at least one case on eu where the current revision was still installed. The repo info from the API contains some error, but the repo is still reported as installed, I guess that the update to the newer revision (with the same version) failed. Also only logging in these cases now. 

TODO 

- Should we change CI such that only for trusted repos revisions are added? 
- For production we should cleanup the temporary clones of the TS repos

Currently running this on this repos. Will need a bit.

martenson:
>It seems that there can be uninstallable revisions that are not due to having the same version as a later revision (e.g. flye from bgruening de24438c9988 where the next installable revision would be 8d4f03b5fe9d

Well that seems like a toolshed bug :/

>graphclust_align_cluster has no installable revisions 

This tool seems pretty broken, the xml is likely invalid but maybe it was not before? 


I think in all these cases we shouldn't modify the lockfile. Maybe create an issue for each and try to track down the problems, if there are any takers.

martenson:
for the latest version I've also encountered this:

```
Traceback (most recent call last):
  File "/Users/marten/devel/git/galaxy_tools/scripts/fix_outdated.py", line 196, in <module>
    fix_outdated(args.lockfile.name, args.toolshed, args.galaxy_url)
  File "/Users/marten/devel/git/galaxy_tools/scripts/fix_outdated.py", line 124, in fix_outdated
    all_revisions = get_all_revisions(toolshed_url, name, owner)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/marten/devel/git/galaxy_tools/scripts/fix_outdated.py", line 67, in get_all_revisions
    assert proc.returncode == 0, f"failed {' '.join(cmd)} in {repo_path}"
           ^^^^^^^^^^^^^^^^^^^^
AssertionError: failed hg update tip in /tmp/repos/toolshed.g2.bx.psu.edu-galaxyp-cardinal_combine
```

the entry looks like this:

```
- name: cardinal_combine
  owner: galaxyp
  revisions:
  - 2b1868ab3724
  - a9a28e46d54a
  - bf06d196e23d
  tool_panel_section_id: proteomics
```

so in my understanding this should just remove the `2b1868ab3724`, but hg fails with the above

bernt-matthias:
So far, for me, deleting the temporary directory for this repo solved the problem (so maybe temporary problem with the remote?). Wondering if we should include stderr, stdout here.

martenson:
the specific repo seems broken :/

```
$ hg clone https://toolshed.g2.bx.psu.edu/repos/galaxyp/cardinal_combine
destination directory: cardinal_combine
requesting all changes
adding changesets
adding manifests
adding file changes
added 21 changesets with 769 changes to 260 files
new changesets 93be1d20b5c3:a9a28e46d54a
updating to branch default
abort: case-folding collision between test-data/cluster_skm.RData and test-data/cluster_skm.rdata
```

bernt-matthias:
Works for me

```
hg clone https://toolshed.g2.bx.psu.edu/repos/galaxyp/cardinal_combine
destination directory: cardinal_combine
requesting all changes
adding changesets
adding manifests
adding file changes
added 21 changesets with 769 changes to 260 files                                                                                                                     
new changesets 93be1d20b5c3:a9a28e46d54a
updating to branch default
218 files updated, 0 files merged, 0 files removed, 0 files unresolved
```

bernt-matthias:
To get an idea I added the changes to the `.lock` files and the log output (`.fix`).

Do not merge yet. Have another idea...

martenson:
I propose showing the changes after run in a separate PR, it makes this PR hard to approach to everyone else

bernt-matthias:
My idea/suggestion is the following: we could have an additional much simpler step that just determines the installed revisions of a given Galaxy instance. Then we just remove everything from the lock files that is not installed. Should be super simple to do.

Should achieve similar results, but with much less resources: This script downloads GBs from the TS and just takes a lot of time -- at least for the first run - I guess subsequent runs will be much faster.

martenson:
@bernt-matthias I understand the urge to optimize but as you said -- to me this is basically a one-off cloning spree triggered only during the first run. Subsequent runs will have one repo to clone at most.

bernt-matthias:
> @bernt-matthias I understand the urge to optimize but, as you said, to me this is basically a one-off cloning spree triggered only during the first run. Subsequent runs will have one repo to clone at most.

Sure, but my main motivation would be that its much simpler, i.e. makes less assumptions on things that are "not well documented" and understood by only a few people.

martenson:
I think that is a fair take.

It likely does not cover the situation where you actually want to update the revision of the repo that did not bump tool version and maybe other corner cases, but likely does majority of the cleanup we want.

bernt-matthias:
I see: the tool version could be extracted from the `guid`. This could then replace `get_all_versions`. How about `get_all_revisions`?

arash77:
> Could be impossible to `get_next` if the list of revisions is incomplete?

Do we need this?
From what I understand, if we have only one installable revision for each version, Galaxy installs the latest revision automatically anyway, right?

bgruening:
I think its here: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/scripts/fix_outdated.py

PR #859, opened 2025-05-11
add vpt segment and extract

nilchia:


bgruening (approved):

bgruening.yaml
```diff
@@ -480,7 +480,14 @@ tools:
   - name: music_manipulate_eset
     owner: bgruening
     tool_panel_section_label: Single-cell
-
+    
+  - name: vpt_extract
+    owner: bgruening
+    tool_panel_section_label: Imaging
```
bgruening, line 486: Just checking, haven't you said you want to put them into single-cell?

nilchia: Thanks a lot! :)
Yes, I copied from the imaging section to single-cell section and forgot to change the label.
Fixed now :+1:

bgruening (approved):

PR #867, opened 2025-05-23
Astro tools: update existing, add 3 new

dsavchenko:


bgruening:
Merci!

dsavchenko:
Is there a delay for deployment of the tools? The ones defined here haven't appeared at usealaxy.eu yet.

bgruening:
Yes, we only deploy over the weekend. If you need it earlier I can trigger ...

dsavchenko:
> Yes, we only deploy over the weekend. If you need it earlier I can trigger ...

Ah, now I remember. No worries, let it deploy as usual

PR #887, opened 2025-06-30
add Graphlan

SantaMcCloud:


bgruening:
You can fix that on IUC by changing the .shed.yml file.

PR #893, opened 2025-07-18
fits2tiff and source-extractor

dsavchenko:


dsavchenko:
@Andrei-EPFL thank you for the new tools

bgruening:
The TS needs 20-30min to update the tool cache.

Thanks everyone!

PR #899, opened 2025-08-01
Add tools required for bachelor project

SaimMomin12:


diverse.yaml
```diff
@@ -47,3 +47,7 @@ tools:
   - name: fbqs_integrator
     owner: maciek
     tool_panel_section_label: Assembly
+
+  - name: nanofilt 
+    owner: leomrtns
+    tool_panel_section_label: Nanopore
```
bgruening, line 53: this will not be autoupdated, so I think you need to add the yml.lock entry as well.

SaimMomin12: Added in [e119984](https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/899/commits/e119984bc994b6a01997f85f0f99ab6b66817652)

PR #906, opened 2025-08-10
add physicell data loader tool suite.

elmbeech:
hi There!

please consider.
these tools are downstream analysis tools for output from the PhysiCell Studio interactive tool, available in the Galaxy Europe (and US) instance.

thank you, Elmar

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -4,79 +4,6 @@ install_resolver_dependencies: true
 install_tool_dependencies: false
 
 tools:
-# AGENT BASED MODELLING
-  - name: pcdl_get_anndata
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
-  - name: pcdl_get_cell_attribute
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
-  - name: pcdl_get_cell_attribute_list
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
-  - name: pcdl_get_cell_df
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
-  - name: pcdl_get_celltype_list
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
-  - name: pcdl_get_conc_attribute
-    owner: iuc
-    tool_panel_section_label: Agent-based Modelling
-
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @elmbeech! Do you find any category where your tools would fit into https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/.schema.yaml#L24 ?

elmbeech:
hi Bjoern : ).

I had a look at the categories.

The closest I see (excluding "Other Tools") is "Single-cell" maybe "Spatial Omics" (but this is a bit too specific).

The whole idea on pcdl is that PhysiCell agent-based modelling output can be transformed into data formats common in the single cell community  (especially data frames, and anndata).
there are some additional command to check on the fly for the software version that generated the output, the substrates and cell types that were modeled, and some basic plotting functions.

If I had to choose from the available categories in the link, I would choose "Single-cell".
But when people hear single cell, they usually think about a different thing thing than agent-based model simulation. 
Taking a look at https://usegalaxy.eu/ , "Single-cell" is under "Genomics Analysis " which is definitely wrong.

Personally, I would do:  ""Miscellaneous Tools / Agent-based Modelling" (with two L in Europe and one L in the US? I leave this up to the admin!)  and add PhysiCell-Studio to this category (in addition to its interactive tool category).

elmbeech:
Hi @bgruening ,

I changed the category to Other Tools.
It is the closest that exists to what the tools are.

Thnak you, Elmar

bgruening:
Ah, I wanted to work on this sorry. It just needed a bit more time to add a new category, but I can do that.

elmbeech:
All well! I know, the weeks are loaded!
Thank you very much, Elmar

bgruening:
Enjoy! And let me know if it makes sense to create some workflows out of it.

Elmar, it might also help - and sorry to give you more and more tasks - to write a small blog post that shows the connection between the IT and the tools. I can then link this blog post up with users - or potential users.

bgruening:
If you want to talk to our single cell community, you can join matrix https://matrix.to/#/#spoc3:matrix.org

They are also here: https://galaxyproject.org/community/sig/singlecell/

>  (I was wondering what you mean with "between information technology and the pcdl tool")

Oh gosh, I just failed hard here :) Sorry.

PR #914, opened 2025-09-08
Add new tb_profiler_collate

pvanheus:
Add the TBProfiler collate tool that combines output from multiple TBProfiler profile runs


pvanheus:
The CI failure is because `AssertionError: bwa_mem_index_builder_data_manager not in toolshed.` which is, in fact, true but beyond the scope of my PR.

PR #1005, opened 2026-03-29
Update ecology.yaml add yolo tools in Computer Vision Ecology section

yvanlebras:
As it seems to me better to do so to have yolo tools on ecology subdomain instead of "loading" the "Imaging" section.


bgruening:
@yvanlebras do we need this PR still?

yvanlebras:
Thank you Björn for tracking it. Seems to me this is not more needed ! Wishing you a very good day!

PR #1021, opened 2026-04-25
Add `parse_image_metadata` tool to imaging.yaml

kostrykin:
Tool: https://github.com/BMCV/galaxy-image-analysis/pull/226/

### DIFF
--- a/imaging.yaml
+++ b/imaging.yaml
@@ -469,3 +469,7 @@ tools:
   - name: cv2_extract_frames
     owner: imgteam
     tool_panel_section_label: "Imaging"
+
+  - name: parse_image_metadata
+    owner: imgteam
+    tool_panel_section_label: "Imaging"


bgruening:
Draft?

kostrykin:
@bgruening I think this tool also needs to be hidden from the tool panel? (but not in the workflow editor)

bgruening:
Why?

bgruening:
I think this is by default for all expression tools.

kostrykin:
> I think this is by default for all expression tools.

Apparently not:

Bildschirmfoto 2026-04-26 um 12 18 41

The tool is displayed in the tool panel also outside of the workflow editor, but cannot be clicked. Nothing happens if you click on it.

bgruening:
This is because it needs to be in the "expression tools" category

PR #1027, opened 2026-05-02
Update ecology.yaml add RLDNE tool

yvanlebras:


yvanlebras:
Thank to you @bgruening , it seems the tool is now on the main toolshed. So it seems to me it is ok to add it to eu!

PR #1030, opened 2026-05-11
feat: Automate tool synchronization from galaxytools repositories

### LINKED ISSUE #1024: feat: Automate tool synchronization from galaxytools repositories
Suggested in https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/1017#pullrequestreview-4147673021

We should implement a generic `sync-tools` mechanism to automate tool updates from various `galaxytools` repositories.

arash77:
## Summary

- Refactor the IUC-only sync workflow and script into a generic `sync-tools` mechanism that supports multiple source repositories
- Add `bgruening/galaxytools` as a second source alongside `galaxyproject/tools-iuc`
- Rename files for clarity: `sync-iuc-tools.yml` → `sync-tools.yml`, `sync-iuc-tools.py` → `sync-tools-repo.py`

Closes #1024

## Changes

- **`.github/workflows/sync-tools.yml`**: New workflow using a matrix strategy to sync from multiple repos (IUC + bgruening). Each source gets its own branch (`sync-{name}-tools`), skip-list key, and last-sync SHA file.
- **`scripts/sync-tools-repo.py`**: Generalized sync script with `--source-repo-url`, `--skip-list-key`, and `--last-sync-sha-file` arguments to work with any galaxytools repository.
- **`scripts/sync-skipped-tools.yml`**: Added `bgruening` key (empty list) alongside existing `iuc` skips.

## Test plan

- [ ] Run the workflow manually with `workflow_dispatch` for both matrix entries
- [ ] Verify the IUC sync still produces the same results as before
- [ ] Verify the bgruening sync discovers tools from `bgruening/galaxytools`
- [ ] Confirm PR creation works with per-source branch names


scripts/sync-skipped-tools.yml
```diff
@@ -36,3 +38,5 @@ skip:
   - qiime_summarize_taxa_through_plots
   - qiime_upgma_cluster
   - qiime_validate_mapping_file
+
+bgruening: []
```
bgruening: For an initial list, please put all tools in here that are in bgruening/galaxytools but are NOT installed on EU

arash77: I've populated the `bgruening` skip list with all 293 tools from `bgruening/galaxytools` that are not currently installed on EU (4f900ea).

arash77: Correction: my initial count of 293 was wrong — it only checked `bgruening.yaml` for already-installed tools, missing 131 tools that are installed in other yaml files (cheminformatics, epigenetics, europe-custom, imaging, rnateam, etc.). After cross-checking all yaml files, the correct skip list has **171 tools** (9a4896d).

bgruening (commented):
Looks good to me.

scripts/sync-skipped-tools.yml
```diff
@@ -0,0 +1,213 @@
...
+  - viennarna_rnalalifold
+  - viennarna_rnalfold
+  - viennarna_rnapaln
+  - viennarna_rnapkplex
+  - viennarna_rnaplex
+  - viennarna_rnaplfold
+  - viennarna_rnaplot
+  - viennarna_rnasnoop
+  - viennarna_rnasubopt
+  - viennarna_rnaup
```
bgruening, line 212: @pavanvidem seems that those tools, assuming the IDs are correct, never made it into the YAML files here, and only where installed manually.

If you want them installed and updated regularly, you need to remove them from this list AND add them into one of the YAML files, for example the bgruening.yaml one.

pavanvidem: Good catch! No wonder why no linting errors as there should be :) I still have to fix the conda recipe to resolve kinwaler segfault but I couln't. I will try again this week or next week. Once we have a proper fix of the tool, I will add it to the yaml file.

pavanvidem: ah ok, they're skipped tools. Ignore the first part of my previous message

bgruening (approved):
Looks solid to me! Thanks @arash77

PR #1033, opened 2026-05-13
20260513 eirene updates

hechth:
added list of new tools from EIRENE, including mzspeclib and mztab validation as well as tabular data manipulation tools


hechth:
wait for https://github.com/bgruening/galaxytools/pull/1866 prior to merging this

PR #1037, opened 2026-05-21
fix(sync): only advance sync SHA pointer when new tools are added

arash77:
This PR fixes the issue where the tools-sync workflow creates empty/SHA-only Pull Requests containing no new tools (as described in https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/1034#issuecomment-4459803973).

By only updating/writing the 'last_sync_sha_file' when new tools are actually found (i.e. 'self.new_tools' is non-empty), we avoid generating unnecessary commits and empty PRs, and ensure that the script will re-evaluate the same source commits in subsequent runs until actual tool updates are processed.

### DIFF
--- a/scripts/sync-tools-repo.py
+++ b/scripts/sync-tools-repo.py
@@ -1029,8 +1029,10 @@ Important:
                     file=sys.stderr,
                 )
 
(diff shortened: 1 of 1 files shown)


bgruening:
Danke!

PR #1038, opened 2026-05-22
Add pecat tool to EU

SaimMomin12:


bgruening:
If you want to be more lazy you can now run this workflow: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/actions/workflows/sync-tools.yml :)

Thanks @SaimMomin12

PR #1045, opened 2026-06-05
Fix bug in fix-lockfile.py preventing new tools from being added

arash77:
Fixes line 28 which checked `if len(locked) == 0:` instead of `if len(locked_tools) == 0:`.

This caused new tools to be silently skipped from the lock file, preventing them from being installed on usegalaxy.eu.

Fixes CI lint failure: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/actions/runs/26823713956/job/79085108349


bgruening:
Thanks! Can you please test run it?

PR #1049, opened 2026-06-09
Fix rate limiting in fix-outdated-tools workflow

arash77:
Fixes HTTP 429 errors from Galaxy Toolshed by:
- Adding 429 to retry codes with 5s initial backoff
- Reducing thread pool workers from 10 to 3
- Adding 0.5s delay between tool requests
- Limiting matrix concurrency to 4 jobs

### DIFF
--- a/scripts/fix_outdated.py
+++ b/scripts/fix_outdated.py
@@ -106,11 +106,12 @@ def fix_uninstallable(lockfile_name, toolshed_url):
                 f"Progress: {i}/{total} tools ({skipped} skipped, {changed} changed)"
             )
 
+        # Brief pause every 50 requests to avoid overwhelming the server
+        if i > 0 and i % 50 == 0:
+            time.sleep(1)
+
         name, owner = tool.get("name"), tool.get("owner")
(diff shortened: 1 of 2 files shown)


scripts/fix_outdated.py
```diff
@@ -105,6 +108,10 @@ def fix_uninstallable(lockfile_name, toolshed_url):
 
         name, owner = tool.get("name"), tool.get("owner")
         current_revisions = set(tool.get("revisions", []))
+
+        # Add delay between requests to avoid rate limiting
+        time.sleep(0.5)
```
bgruening: Not sure if this is wise. 4000 tools, many duplicated ... this will maybe take too long?

arash77: Should we remove the delay completly as we have a retry with backoff here or just reduce it?

bgruening: Whatever works and is not completely slow. Maybe just wait every 100 requests or so?

arash77: I added a 1s pause every 50 requests.

bgruening:
Danke.

PR #1083, opened 2026-07-03
Combine tool-maintenance workflows into a single weekly pipeline

arash77:
## What

Consolidates the three tool-maintenance workflows — `sync-tools`, `update-trusted`, and `fix-outdated-tools` — into a single ordered **weekly-tools-update** pipeline so that, before the Sunday installs on usegalaxy.eu:

1. **Sync** — new tools are added (from `galaxyproject/tools-iuc` and `bgruening/galaxytools`)
2. **Update** — new installable revisions are added for trusted tool owners
3. **Deprecate** — revisions no longer installable from the ToolShed are removed

Running the three steps in one pipeline (in that order) prevents the races from independent schedules: new tools land before their revisions are fetched, and outdated revisions are removed last.

## Pipeline

```
get-targets → sync (matrix, 2 legs) → update → deprecate (matrix, 1 leg per lockfile) → assemble-pr
```

Each step hands off via artifacts, and `assemble-pr` opens a single combined PR. `deprecate` is a per-lockfile matrix (mirroring `fix-outdated-tools.yml`) so each lockfile gets its own green/red leg in the Actions UI; a failing leg doesn't discard the run's sync/update work. Runs every Friday 06:00 UTC.

## Changes

- **New: `weekly-tools-update.yml`** — the combined pipeline above.
- **New: `scripts/sync-targets.json`** — single source of truth for the sync matrix, consumed by both the weekly and the per-repo `sync-tools` workflows.
- **`Makefile`** — adds `sync`, `update`, and `deprecate` targets. `sync` takes a single target via `NAME=`/`REPO=`/`YAML=` or all targets from `sync-targets.json`.
- **`sync-tools.yml`** — matrix driven from `sync-targets.json`, runs via `make sync`, validates the synced yaml directly (not `make pr_check`, which diffs `origin/master` and sees no uncommitted changes; and not `make <file>.lint`, whose `fix-lockfile.py` would regenerate the lockfile that `update` owns).
- **`update-trusted.yml`** — uses `make update`.
- Schedules removed from the three individual workflows in favour of the combined weekly run; they remain available for manual `workflow_dispatch`.
- Bumps `actions/checkout`, `actions/setup-python`, and `astral-sh/setup-uv` across the touched workflows.

## Notes / follow-ups

A pre-PR review surfaced two issues worth addressing (can be follow-ups, flagged for transparency):

- **`make pr_check` after `make update` makes the toolshed-existence check a no-op for synced tools** — `make update` regenerates lockfiles first, so `pr-check.py`'s `new_tools = yml − yml_lock` is empty and never checks that newly-synced tools exist in the toolshed. The per-repo `sync-tools.yml` validates before lockfile regen and still catches this.
- **A single transient toolshed fetch failure for one new tool fails `make update`** (no `continue-on-error`) and discards the whole run's sync+update work, since `update-tool.py --without` swallows the fetch error leaving the tool unpinned and `identify-unpinned.py` then exits 1.

## Verification

- Rebased onto the latest `upstream/master` (resolves cleanly).
- All touched workflow files parse as valid YAML and pass `actionlint`; `zizmor` is clean at the default persona.
- Triggered `workflow_dispatch` runs on the fork: all jobs green (2 sync legs, 34 deprecate legs, assemble-pr), and the generated PR body renders correctly.


arash77:
Preview of a live run + the generated PR body: https://github.com/arash77/usegalaxy-eu-tools/pull/7

bgruening (approved):
thanks @arash77!

PR #1090, opened 2026-07-16
Organize tools into appropriate sections

pavanvidem:
Added new tools under Digital Humanities and Metabolomics sections while removing them from Other Tools section.


Sch-Da:
Thanks a lot, @pavanvidem !

PR #1098, opened 2026-07-27
Add add_input_name_as_column tool to Text Manipulation

bgruening:
@arash77 I think this was not added because it has a  different owner than IUC? Can this be? However, it is now part of IUC. This can happen when tools are moved later to IUC, so IUC 'adopts' those tools and maintains them.

Can you bot also catch those?

Or is there a different reason this tool is not added?

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -6664,6 +6664,10 @@ tools:
     owner: devteam
     tool_panel_section_label: Text Manipulation
 
+  - name: add_input_name_as_column
+    owner: mvdbeek
+    tool_panel_section_label: Text Manipulation
+
   - name: column_order_header_sort
     owner: iuc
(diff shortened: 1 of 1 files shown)


arash77:
@bgruening It would have been added via the catchup option, but we didn't run it for IUC. I will run it now to add anything we missed so far.

bgruening:
Thanks for the explanation! Do we need to run this option by default. It might happen more regularly that people add tools "for adaption" to IUC with a different owner.

arash77:
> Thanks for the explanation! Do we need to run this option by default. It might happen more regularly that people add tools "for adaption" to IUC with a different owner.

That wouldn't be necessary, as the script should find them. `add_input_name_as_column` hasn't been added because it was added before we adopted this script (in 2020), and for IUC we didn't run it with catchup to get the old tools.

PR #1104, opened 2026-07-31
Add deacon

pvanheus:
This adds the [deacon](https://github.com/bede/deacon) tools for filtering and de-hosting reads. Wrappers were recently [added to tools-iuc](https://github.com/galaxyproject/tools-iuc/pull/8247).

(I am not clear if the individual tools need to get mentioned or if they'll all be pulled in by the mention of deacon)

### DIFF
--- a/tools_iuc.yaml
+++ b/tools_iuc.yaml
@@ -2030,6 +2030,10 @@ tools:
     owner: iuc
     tool_panel_section_label: FASTA/FASTQ
 
+  - name: deacon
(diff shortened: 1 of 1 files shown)


bgruening:
@pvanheus https://github.com/galaxyproject/tools-iuc/pull/8247/changes#r3695333752

And please also add the data manager when the tool is updated.

bgruening:
I think this is now installed correct?

PR #1109, opened 2026-08-05
new tool: Create a ZIP file from a collection

rmassei:
Adding a new tool to ZIP selected files from a collection.
All files can be zipped or just the ones with a specific file extension

### DIFF
--- a/ufz.yaml
+++ b/ufz.yaml
@@ -64,3 +64,7 @@ tools:
   - name: omero_dataset_to_plate
     owner: ufz
     tool_panel_section_label: Imaging
+  - name: zipping
+    owner: ufz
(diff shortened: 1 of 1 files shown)


bgruening:
Is that for download a collection?

rmassei:
No, it creates a ZIP file from the datasets in a collection, with the option to filter by data type. For example, if a collection has TIFF files mixed with TXT files and so on, the user can create a ZIP with just the TIFF files.

This can be useful for some tools, such as CellProfiler, which accept images stored in a ZIP file. Otherwise, the workflow will be broken since user need to download the collection and then upload it again.

bgruening:
As per offline discussion, we will wait here

PR #1115, opened 2026-08-11
Show new tools inline and collapse removed revisions in the weekly PR body

arash77:
Follow-up to @bgruening's comment on #1113:

> > Removed not-installable revisions
>
> should be hidden and the new tools should be shown I guess?

The weekly PR body currently does the opposite: the sync reports are tucked into a collapsed `<details>`, while the removed-revisions table is dumped inline — in #1113 that was 71 rows pushing everything else off screen.

This swaps the two:

- **New tools** are inlined under a `### Sync report: <name>` heading. Each report's own `# Tools Sync Report` H1, trailing `---`, and redundant "🤖 automatically generated by the `sync-tools-repo` workflow" footer are stripped, and its `##` headings demoted to `####`, so the reports nest under the PR body's sections instead of restarting the heading hierarchy.
- **Removed not-installable revisions** keeps its heading and explanatory sentence visible, but the table itself moves into `<details><summary>Show N removed revisions</summary>`.

Only the PR body formatting changes; no change to what is synced, updated, or deprecated.

### DIFF
--- a/.github/workflows/weekly-tools-update.yml
+++ b/.github/workflows/weekly-tools-update.yml
@@ -340,15 +340,19 @@ jobs:
             echo "Synced from \`galaxyproject/tools-iuc\` and \`bgruening/galaxytools\`."
             echo ""
             if [ "$new_tools" -gt 0 ] && ls sync-report-*.md >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


bgruening (approved):

bgruening:
Thanks

PR #1133, opened 2026-09-02
Add mirtrace_qc and mirtrace_trace tools

pavanvidem:
I don't know exactly why they were not caught in the auto update workflow. Maybe because I initally added wrong owner in the .shed.yml


bgruening:
what user was it?

pavanvidem:
I initially added as iuc but it should be bgreuning. Is that what  you asked for?

PR #1134, opened 2026-09-03
Add langchain_text_splitters tool to Digital Humanities section in bgruening.yaml

IvoLeist:
Corresponding merged PR:
https://github.com/bgruening/galaxytools/pull/1947


bgruening:
For bgruening and IUC this is now happening automatically. For all other repos we need to add it manually as you did here.

pavanvidem:
I remember this but was looking at the wrong WF.

PR #1136, opened 2026-09-09
Add simorgh tool to Machine Learning section

nilchia:
requires this https://github.com/bgruening/galaxytools/pull/1964 to be deployed.


SaimMomin12 (approved):

pavanvidem:
BTW, that should happen automatically on Friday, but it might put it into the wrong section ;)

nilchia:
Yes but I want to put the link in the manuscript so they can submit the paper to the journal. Otherwise I would've waited :)

